PR #38, opened 2017-08-25
Pfcli unify state queries

pronebird:
`get_anchors`, `get_rules` and `get_states` have absolutely the same source code. This PR refactors this and unifies firewall state queries into one function `pfcli::query_state(anchor_name, query_kind)`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/38)

### DIFF
--- a/tests/anchors.rs
+++ b/tests/anchors.rs
@@ -26,7 +26,7 @@ test!(add_filter_anchor {
 
     assert_matches!(pf.add_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
 
-    let anchors = pfcli::get_anchors().unwrap();
+    let anchors = pfcli::get_anchors(None).unwrap();
     assert!(anchors.contains(&anchor_name));
 
     assert_matches!(
@@ -43,7 +43,7 @@ test!(remove_filter_anchor {
     assert_matches!(pf.add_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
     assert_matches!(pf.remove_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
 
-    let anchors = pfcli::get_anchors().unwrap();
+    let anchors = pfcli::get_anchors(None).unwrap();
     assert!(!anchors.contains(&anchor_name));
 
     assert_matches!(
--- a/tests/pfctl_test/src/pfcli.rs
+++ b/tests/pfctl_test/src/pfcli.rs
@@ -1,4 +1,3 @@
-use std::ffi::OsStr;
 use std::process::Command;
 use std::str;
 
@@ -58,25 +57,12 @@ pub fn disable_firewall() -> Result<()> {
     Ok(())
 }
 
-pub fn get_anchors() -> Result<Vec<String>> {
-    let output = get_command()
-        .arg("-s")
-        .arg("Anchors")
-        .output()
-        .chain_err(|| "Failed to run pfctl")?;
-    let output = str_from_stdout(&output.stdout)?;
-    let names = output
-        .lines()
-        .map(|x| x.trim().to_owned())
-        .collect();
-    Ok(names)
-}
-
-pub fn get_rules<S: AsRef<OsStr>>(anchor_name: S) -> Result<Vec<String>> {
+fn get_rules_internal(anchor_name: &str, param_kind: &str) -> Result<Vec<String>> {
     let output = get_command()
         .arg("-a")
-        .arg(anchor_name.as_ref())
-        .arg("-sr")
+        .arg(anchor_name)
+        .arg("-s")
+        .arg(param_kind)
         .output()
         .chain_err(|| "Failed to run pfctl")?;
     let output = str_from_stdout(&output.stdout)?;
@@ -87,20 +73,36 @@ pub fn get_rules<S: AsRef<OsStr>>(anchor_name: S) -> Result<Vec<String>> {
     Ok(rules)
 }
 
-pub fn get_states<S: AsRef<OsStr>>(anchor_name: S) -> Result<Vec<String>> {
+/// List anchors.
+/// Pass parent anchor's name to obtain nested anchors.
+/// Otherwise, pass None to obtain anchors from main ruleset.
+pub fn get_anchors(parent_anchor: Option<&str>) -> Result<Vec<String>> {
+    get_rules_internal(parent_anchor.unwrap_or("*"), "Anchors")
+}
+
+/// Get filter rules in anchor
+pub fn get_rules(anchor_name: &str) -> Result<Vec<String>> {
+    get_rules_internal(anchor_name, "rules")
+}
+
+/// Get nat rules in anchor
+pub fn get_nat_rules(anchor_name: &str) -> Result<Vec<String>> {
+    get_rules_internal(anchor_name, "nat")
+}
+
+/// Get global table of states
+pub fn get_all_states() -> Result<Vec<String>> {
     let output = get_command()
-        .arg("-a")
-        .arg(anchor_name.as_ref())
         .arg("-s")
         .arg("states")
         .output()
         .chain_err(|| "Failed to run pfctl")?;
     let output = str_from_stdout(&output.stdout)?;
-    let rules = output
+    let states = output
         .lines()
         .map(|x| x.trim().to_owned())
         .collect();
-    Ok(rules)
+    Ok(states)
 }
 
 
@@ -123,11 +125,11 @@ impl From<FlushOptions> for &'static str {
     }
 }
 
-pub fn flush_rules<S: AsRef<OsStr>>(anchor_name: S, options: FlushOptions) -> Result<()> {
+pub fn flush_rules(anchor_name: &str, options: FlushOptions) -> Result<()> {
     let flush_arg: &'static str = options.into();
     let output = get_command()
         .arg("-a")
-        .arg(anchor_name.as_ref())
+        .arg(anchor_name)
         .arg("-F")
         .arg(flush_arg)
         .output()
--- a/tests/states.rs
+++ b/tests/states.rs
@@ -52,13 +52,13 @@ test!(reset_ipv4_states_by_anchor {
     send_udp_packet(sender_addr, server_addr);
 
     assert_matches!(
-        pfcli::get_states(ANCHOR_NAME),
+        pfcli::get_all_states(),
         Ok(ref v) if v == &["ALL udp 127.0.0.1:1338 -> 127.0.0.1:1337       SINGLE:NO_TRAFFIC",
                             "ALL udp 127.0.0.1:1337 <- 127.0.0.1:1338       NO_TRAFFIC:SINGLE"]
     );
     assert_matches!(pf.clear_states(ANCHOR_NAME, pfctl::AnchorKind::Filter), Ok(2));
     assert_matches!(
-        pfcli::get_states(ANCHOR_NAME),
+        pfcli::get_all_states(),
         Ok(ref v) if v.len() == 0
     );
 });
@@ -73,13 +73,13 @@ test!(reset_ipv6_states_by_anchor {
     send_udp_packet(sender_addr, server_addr);
 
     assert_matches!(
-        pfcli::get_states(ANCHOR_NAME),
+        pfcli::get_all_states(),
         Ok(ref v) if v == &["ALL udp ::1[1338] -> ::1[1337]       SINGLE:NO_TRAFFIC",
                             "ALL udp ::1[1337] <- ::1[1338]       NO_TRAFFIC:SINGLE"]
     );
     assert_matches!(pf.clear_states(ANCHOR_NAME, pfctl::AnchorKind::Filter), Ok(2));
     assert_matches!(
-        pfcli::get_states(ANCHOR_NAME),
+        pfcli::get_all_states(),
         Ok(ref v) if v.len() == 0
     );
 });
--- a/tests/states.rs
+++ b/tests/states.rs
@@ -11,6 +11,14 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket};
 
 static ANCHOR_NAME: &'static str = "pfctl-rs.integration.testing.states";
 
+fn contains_subset(haystack: &Vec<String>, subset: &[&str]) -> bool {
+    subset.iter().all(|&state| haystack.contains(&state.to_owned()))
+}
+
+fn not_contains_subset(haystack: &Vec<String>, subset: &[&str]) -> bool {
+    subset.iter().all(|&state| !haystack.contains(&state.to_owned()))
+}
+
 fn send_udp_packet(sender: SocketAddr, recepient: SocketAddr) {
     UdpSocket::bind(sender)
         .unwrap()
@@ -51,15 +59,19 @@ test!(reset_ipv4_states_by_anchor {
     pf.set_rules(ANCHOR_NAME, &[rule_builder(server_addr)]).unwrap();
     send_udp_packet(sender_addr, server_addr);
 
+    let expected_states = [
+        "ALL udp 127.0.0.1:1338 -> 127.0.0.1:1337       SINGLE:NO_TRAFFIC",
+        "ALL udp 127.0.0.1:1337 <- 127.0.0.1:1338       NO_TRAFFIC:SINGLE"
+    ];
+
     assert_matches!(
         pfcli::get_all_states(),
-        Ok(ref v) if v == &["ALL udp 127.0.0.1:1338 -> 127.0.0.1:1337       SINGLE:NO_TRAFFIC",
-                            "ALL udp 127.0.0.1:1337 <- 127.0.0.1:1338       NO_TRAFFIC:SINGLE"]
+        Ok(ref v) if contains_subset(v, &expected_states)
     );
     assert_matches!(pf.clear_states(ANCHOR_NAME, pfctl::AnchorKind::Filter), Ok(2));
     assert_matches!(
         pfcli::get_all_states(),
-        Ok(ref v) if v.len() == 0
+        Ok(ref v) if not_contains_subset(v, &expected_states)
     );
 });
 
@@ -72,14 +84,18 @@ test!(reset_ipv6_states_by_anchor {
     pf.set_rules(ANCHOR_NAME, &[rule_builder(server_addr)]).unwrap();
     send_udp_packet(sender_addr, server_addr);
 
+    let expected_states = [
+        "ALL udp ::1[1338] -> ::1[1337]       SINGLE:NO_TRAFFIC",
+        "ALL udp ::1[1337] <- ::1[1338]       NO_TRAFFIC:SINGLE"
+    ];
+
     assert_matches!(
         pfcli::get_all_states(),
-        Ok(ref v) if v == &["ALL udp ::1[1338] -> ::1[1337]       SINGLE:NO_TRAFFIC",
-                            "ALL udp ::1[1337] <- ::1[1338]       NO_TRAFFIC:SINGLE"]
+        Ok(ref v) if contains_subset(v, &expected_states)
     );
     assert_matches!(pf.clear_states(ANCHOR_NAME, pfctl::AnchorKind::Filter), Ok(2));
     assert_matches!(
         pfcli::get_all_states(),
-        Ok(ref v) if v.len() == 0
+        Ok(ref v) if not_contains_subset(v, &expected_states)
     );
 });


pronebird:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/pfctl_test/src/pfcli.rs, line 74 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsPA52YGxilvk-_tbFt:-KsPA52YGxilvk-_tbFu:b-b34j0e) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L74)):*
> ```Rust
> 
> /// This is a constant value that represents main ruleset in PF
> pub static MAIN_RULESET: &'static str = "*";
> ```

The reason why this exists is:

1. Calling `query_state("*", QueryKind::Anchors)` returns anchors from main ruleset
2. Calling `query_state("my_anchor", QueryKind::Anchors)` returns sub-anchors from the given anchor

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 68 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ311xH6-OZhFsWUke:-KsZ311yaB99Z7L1PgrL:b-jaica7) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L68)):*
> ```Rust
>     Rules,
>     /// Redirect rules
>     Nat,
> ```

I'm very confused by the fact that "filter rules" are referred to as "rules" but "redirect rules" as "nat". They are both equally much rules. I see it's because how `pfctl` wants it on the command line. But for someone coming at PF from a code point of view this is very strange. Can't the variants be called `FilterRules` and `RedirectRules` instead to make it a bit consistent with what PF usually call them?

---

*[tests/pfctl_test/src/pfcli.rs, line 87 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ2NxWJ2JqeI-AHFa9:-KsZ2NxX-hj64p08QK6f:b-a2o79d) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L87)):*
> ```Rust
> }
> 
> pub fn query_state<S: AsRef<OsStr>>(anchor_name: S, kind: QueryKind) -> Result<Vec<String>> {
> ```

I feel the new name is strange. Both parts of it actually. I think "get" was better than "query", since the goal is to get a list.
I also don't understand where "state" comes from? I thought all of anchors, filter rules and redirect rules were referred to as "rules", only getting the list of states sticks out from that pattern, right?

Since getting these lists happen to share a lot of code, but not very much meaning, maybe it's best to expose separate methods, but use a common helper function to reuse code? Like, the actual implementation that is this method can be a private method like
```rust
/// the name `modifier` comes from what the pfctl man page calls the -s argument.
/// The `parameters_internal` is just something I made up, with inspiration from the `pfctl` man page.
get_parameters_internal<...>(anchor: S, modifier: &str) -> Result<Vec<String>>`
```

And then expose these public ones, that use `get_parameters_internal` internally:
```rust
/// Get all anchors. Here the argument is called "parent" because that makes sense when getting anchors IMO.
/// I made it optional, so the impl can automatically default to the `MAIN_RULESET` nicely.
pub fn get_anchors(parent: Option<S>) -> Result<Vec<String>> {}

/// Both types of rules can share one method. Because they are are tightly coupled, plus there is already a nice non-test enum for describing filter/redirect. Nice to use code from the main crate IMO.
pub fn get_rules(anchor: S, kind: pfctl::RulesetKind) -> Result<Vec<String>> {}


pub fn get_states(anchor: S) -> Result<Vec<String>> {}
```

With this solution you would not even need `QueryKind`. I'm a bit hesitant to reimplement too much just for the tests, so better to not need so many structs and enums representing kind of what we already have in the main crate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 68 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ311xH6-OZhFsWUke:-Ksbyt6dT33AeobrXc2j:bax864q) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L68)):*

I'm very confused by the fact that "filter rules" are referred to as "rules" but "redirect rules" as "nat". They are both equally much rules. I see it's because how `pfctl` wants it on the command line. But for someone coming at PF from a code point of view this is very strange. Can't the variants be called `FilterRules` and `RedirectRules` instead to make it a bit consistent with what PF usually call them?
</blockquote></details>

Not only you, actually terminology even in PF is quite obscure. Excerpt from `man pfctl`:

```
 -s modifier
             Show the filter parameters specified by modifier (may be abbrevi-
             ated):

             -s nat         Show the currently loaded NAT rules.
             -s queue       Show the currently loaded queue rules.  When used
                            together with -v, per-queue statistics are also
                            shown.  When used together with -v -v, pfctl will
                            loop and show updated queue statistics every five
                            seconds, including measured bandwidth and packets
                            per second.
             -s rules       Show the currently loaded filter rules.  When used
                            together with -v, the per-rule statistics (number
                            of evaluations, packets and bytes) are also shown.
                            Note that the ``skip step'' optimization done
                            automatically by the kernel will skip evaluation
                            of rules where possible.  Packets passed state-
                            fully are counted in the rule that created the
                            state (even though the rule isn't evaluated more
                            than once for the entire connection).
             -s Anchors     Show the currently loaded anchors directly
                            attached to the main ruleset.  If -a anchor is
                            specified as well, the anchors loaded directly
                            below the given anchor are shown instead.  If -v
                            is specified, all anchors attached under the tar-
                            get anchor will be displayed recursively.
             -s states      Show the contents of the state table.
             -s Sources     Show the contents of the source tracking table.
             -s info        Show filter information (statistics and counters).
                            When used together with -v, source tracking sta-
                            tistics are also shown.
             -s References  Show pf-enable reference statistics (pid/name of
                            enabler, token, timestamp).

...

```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 68 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ311xH6-OZhFsWUke:-KsbzT42x4PTBhhmfROd:bmmtf3c) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L68)):*

Not only you, actually terminology even in PF is quite obscure. Excerpt from `man pfctl`:

```
 -s modifier
             Show the filter parameters specified by modifier (may be abbrevi-
             ated):

             -s nat         Show the currently loaded NAT rules.
             -s queue       Show the currently loaded queue rules.  When used
                            together with -v, per-queue statistics are also
                            shown.  When used together with -v -v, pfctl will
                            loop and show updated queue statistics every five
                            seconds, including measured bandwidth and packets
                            per second.
             -s rules       Show the currently loaded filter rules.  When used
                            together with -v, the per-rule statistics (number
                            of evaluations, packets and bytes) are also shown.
                            Note that the ``skip step'' optimization done
                            automatically by the kernel will skip evaluation
                            of rules where possible.  Packets passed state-
                            fully are counted in the rule that created the
                            state (even though the rule isn't evaluated more
                            than once for the entire connection).
             -s Anchors     Show the currently loaded anchors directly
                            attached to the main ruleset.  If -a anchor is
                            specified as well, the anchors loaded directly
                            below the given anchor are shown instead.  If -v
                            is specified, all anchors attached under the tar-
                            get anchor will be displayed recursively.
             -s states      Show the contents of the state table.
             -s Sources     Show the contents of the source tracking table.
             -s info        Show filter information (statistics and counters).
                            When used together with -v, source tracking sta-
                            tistics are also shown.
             -s References  Show pf-enable reference statistics (pid/name of
                            enabler, token, timestamp).

...

```
</blockquote></details>

Maybe I should change the comment. It could be that `-s nat` shows all of kinds of nat rules, and there are many of them including redirect rules?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 87 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ2NxWJ2JqeI-AHFa9:-Ksc-i0Y_Cjvqm3dR9JB:b4s6o4p) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L87)):*

I feel the new name is strange. Both parts of it actually. I think "get" was better than "query", since the goal is to get a list.
I also don't understand where "state" comes from? I thought all of anchors, filter rules and redirect rules were referred to as "rules", only getting the list of states sticks out from that pattern, right?

Since getting these lists happen to share a lot of code, but not very much meaning, maybe it's best to expose separate methods, but use a common helper function to reuse code? Like, the actual implementation that is this method can be a private method like
```rust
/// the name `modifier` comes from what the pfctl man page calls the -s argument.
/// The `parameters_internal` is just something I made up, with inspiration from the `pfctl` man page.
get_parameters_internal<...>(anchor: S, modifier: &str) -> Result<Vec<String>>`
```

And then expose these public ones, that use `get_parameters_internal` internally:
```rust
/// Get all anchors. Here the argument is called "parent" because that makes sense when getting anchors IMO.
/// I made it optional, so the impl can automatically default to the `MAIN_RULESET` nicely.
pub fn get_anchors(parent: Option<S>) -> Result<Vec<String>> {}

/// Both types of rules can share one method. Because they are are tightly coupled, plus there is already a nice non-test enum for describing filter/redirect. Nice to use code from the main crate IMO.
pub fn get_rules(anchor: S, kind: pfctl::RulesetKind) -> Result<Vec<String>> {}


pub fn get_states(anchor: S) -> Result<Vec<String>> {}
```

With this solution you would not even need `QueryKind`. I'm a bit hesitant to reimplement too much just for the tests, so better to not need so many structs and enums representing kind of what we already have in the main crate.
</blockquote></details>

I think `QueryKind` is a clear way to manage parameters passed to `-s`, `From` trait is quite handy here. Also, if we decide to add `Interfaces` or something in future, it's easy to extend `match` and obtain filters from enum cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 87 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ2NxWJ2JqeI-AHFa9:-Ksc0o4qnAOrjuVFDYFO:b-qq3c68) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L87)):*

I think `QueryKind` is a clear way to manage parameters passed to `-s`, `From` trait is quite handy here. Also, if we decide to add `Interfaces` or something in future, it's easy to extend `match` and obtain filters from enum cases.
</blockquote></details>

I agree that we could add convenience methods to make it easier to obtain needed information.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 87 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ2NxWJ2JqeI-AHFa9:-Ksc4uV6q0isLx9DmOFV:b-u8n9v8) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L87)):*

I agree that we could add convenience methods to make it easier to obtain needed information.
</blockquote></details>

If we add convenience methods then `QueryKind` has no use, because the impls of those convenience methods would simply be
```rust
pub fn get_anchors(parent: Option<S>) -> Result<Vec<String>> {
    get_parameters_internal(parent.unwrap_or("*"), "Anchors")
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Reviewed 1 of 4 files at r2.
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/states.rs, line 57 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaqRxGvIAI-5it-zj:-KscaqRxGvIAI-5it-zk:bubumez) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/states.rs#L57)):*
> ```Rust
>         pfcli::get_all_states(),
>         Ok(ref v) if v == &["ALL udp 127.0.0.1:1338 -> 127.0.0.1:1337       SINGLE:NO_TRAFFIC",
>                             "ALL udp 127.0.0.1:1337 <- 127.0.0.1:1338       NO_TRAFFIC:SINGLE"]
> ```

Won't it be very hard to guarantee that these are the only states in the global PF? Any other state at all would break this, would it not? So maybe check if the two states you want are present in `v`, not that `v` is exactly this slice.

---

*[tests/pfctl_test/src/pfcli.rs, line 94 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaZq1h76rwtcCrwnZ:-KscaZq1h76rwtcCrwn_:b-uwy8cg) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/pfctl_test/src/pfcli.rs#L94)):*
> ```Rust
> 
> /// Get global table of states
> pub fn get_all_states() -> Result<Vec<String>> {
> ```

Does the command break if you send `-a *` or similar to it? Otherwise you could make use of `get_rules_internal` here as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Reviewed 1 of 4 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/states.rs, line 57 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaqRxGvIAI-5it-zj:-KscgJwCJCAoxHeOUCHy:bdaflz8) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/states.rs#L57)):*

Won't it be very hard to guarantee that these are the only states in the global PF? Any other state at all would break this, would it not? So maybe check if the two states you want are present in `v`, not that `v` is exactly this slice.
</blockquote></details>

We spoke about that, and basically this should not happen on Travis.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 94 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaZq1h76rwtcCrwnZ:-KscgQ6VgjjE4uV1xJ_U:b-c9peto) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/pfctl_test/src/pfcli.rs#L94)):*

Does the command break if you send `-a *` or similar to it? Otherwise you could make use of `get_rules_internal` here as well.
</blockquote></details>

It ignores `-a`, but I felt weird about reusing command with the ignored parameter... what it they fix it and it'll work tomorrow...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 1 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 87 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KsZ2NxWJ2JqeI-AHFa9:-Kscgc_bpRbV30OmQS3M:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/369b63b8d64be7c1b8fa8383544626b476a63a0a/tests/pfctl_test/src/pfcli.rs#L87)):*

If we add convenience methods then `QueryKind` has no use, because the impls of those convenience methods would simply be
```rust
pub fn get_anchors(parent: Option<S>) -> Result<Vec<String>> {
    get_parameters_internal(parent.unwrap_or("*"), "Anchors")
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 1 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/pfctl_test/src/pfcli.rs, line 94 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaZq1h76rwtcCrwnZ:-Kscgh4xOZLF5OLKsJoQ:b-5bu3hk) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/pfctl_test/src/pfcli.rs#L94)):*

It ignores `-a`, but I felt weird about reusing command with the ignored parameter... what it they fix it and it'll work tomorrow...
</blockquote></details>

what it = what if

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Reviewed 1 of 4 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/states.rs, line 57 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaqRxGvIAI-5it-zj:-Ksch6AvKQcRm7HPjdXJ:b-va2vjz) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/states.rs#L57)):*

We spoke about that, and basically this should not happen on Travis.
</blockquote></details>

But tests are not written exclusively for Travis. One must be able to execute them locally as well. And I might have hundreds of ongoing connections (and thus states) on my local machine)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/states.rs, line 57 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaqRxGvIAI-5it-zj:-KschKoLFpFZEAyNg9S_:b4rko2z) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/states.rs#L57)):*

But tests are not written exclusively for Travis. One must be able to execute them locally as well. And I might have hundreds of ongoing connections (and thus states) on my local machine)
</blockquote></details>

Even for Travis we have no idea if their machines have zero state by default. Maybe they do today, but what about tomorrow? Feels like a fragile part.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/states.rs, line 57 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscaqRxGvIAI-5it-zj:-Ksczs9fww8BU3CIRylv:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/004f74de9befb7a87776ab61b8e80ca65371d903/tests/states.rs#L57)):*

Even for Travis we have no idea if their machines have zero state by default. Maybe they do today, but what about tomorrow? Feels like a fragile part.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 90 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscztqntiNuWrYmOkd3:-KscztqntiNuWrYmOkd4:buph82d) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L90)):*
> ```Rust
>         "ALL udp ::1[1338] -> ::1[1337]       SINGLE:NO_TRAFFIC",
>         "ALL udp ::1[1337] <- ::1[1338]       NO_TRAFFIC:SINGLE"]
>         .iter().map(|&s| s.to_owned()).collect();
> ```

I chose `iter().map().collect()` over `.to_owned()` for each element because I like test data to be easily swappable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 90 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscztqntiNuWrYmOkd3:-Ksd-54G-rYymQEbe-eE:b-wj7sy4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L90)):*

I chose `iter().map().collect()` over `.to_owned()` for each element because I like test data to be easily swappable.
</blockquote></details>

This is probably less relevant now after these tests were made, but it's still easier to read conditions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 2 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd0VNU9kz8YqDgxLbu:bbuw6gr) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*
> ```Rust
> 
> fn not_contains_subset(haystack: &Vec<String>, subset: &Vec<String>) -> bool {
>     subset.iter().all(|state| !haystack.contains(state))
> ```

Neither of these methods need to modify the data, just look at it. So a more correct type to accept would be `&[&str]` instead of `Vec<...>`. Just sending slices around is better when the intention is just to look at the data. Then you don't have to go the extra step by mapping to `.to_owned` etc for the test data further down.

Also I think it's a bit overkill to create two helper methods when they are just oneliners.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 90 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscztqntiNuWrYmOkd3:-Ksd14rdpbDtttbDdXt3:b-f0bobm) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L90)):*

This is probably less relevant now after these tests were made, but it's still easier to read conditions.
</blockquote></details>

If you just did the oneliner check here instead of handing it off to a helper function you would not need the extra step in converting them to owned strings I think.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd1LTiQPF-WYisVlXV:b-ap6zcx) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

Neither of these methods need to modify the data, just look at it. So a more correct type to accept would be `&[&str]` instead of `Vec<...>`. Just sending slices around is better when the intention is just to look at the data. Then you don't have to go the extra step by mapping to `.to_owned` etc for the test data further down.

Also I think it's a bit overkill to create two helper methods when they are just oneliners.
</blockquote></details>

`contains` is still restricted to `String` so we would have to convert to `String` in `contains()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd2FXoBgxvXcPrzJPd:bwnuwpr) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

`contains` is still restricted to `String` so we would have to convert to `String` in `contains()`
</blockquote></details>

can be:

```
subset.iter().all(|&state| haystack.contains(&state.to_owned()))
```

No problems if we copy around a little bit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 90 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-KscztqntiNuWrYmOkd3:-Ksd2SpJMSjrnN3o_bSf:bqa3ll1) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L90)):*

If you just did the oneliner check here instead of handing it off to a helper function you would not need the extra step in converting them to owned strings I think.
</blockquote></details>

`Vec<String>::contains(search: &String)`, I still need to convert `&str` to owned string otherwise it won't work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd2b461FhAHcAb4SGw:b-f25zni) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

can be:

```
subset.iter().all(|&state| haystack.contains(&state.to_owned()))
```

No problems if we copy around a little bit.
</blockquote></details>

Sounds silly that we would have to allocate the string on the heap, just to compare it. But I can't come up with another solution on the top of my head at the moment either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd2l1wKE_00Ln04F17:bbsn9cv) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

Sounds silly that we would have to allocate the string on the heap, just to compare it. But I can't come up with another solution on the top of my head at the moment either.
</blockquote></details>

I agree, I thought it should just work but there is an issue regarding `Vec::contains()` flexibility and attempts to relax the requirements.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd2ktbkeAUyYbP4XP8:b-cdrxfj) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

I agree, I thought it should just work but there is an issue regarding `Vec::contains()` flexibility and attempts to relax the requirements.
</blockquote></details>

Or, well, the obvious is to do it the other way around, map `v` into `&[&str]` of course. Not less code, but less allocation. Not that I mean allocation is a problem in the test, reducing code is more important.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd46h88_Oi_o71mUNF:b4fommr) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

Or, well, the obvious is to do it the other way around, map `v` into `&[&str]` of course. Not less code, but less allocation. Not that I mean allocation is a problem in the test, reducing code is more important.
</blockquote></details>

I meant both `haystack` and `subset` should probably be slices and not `Vec`s. But I mark all of this as accepted now. Too much time spent on polishing tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd5q4vGO59TPsSpMtC:bx081ts) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

I meant both `haystack` and `subset` should probably be slices and not `Vec`s. But I mark all of this as accepted now. Too much time spent on polishing tests.
</blockquote></details>

One of recent revisions changes subset to `&[&str]` slice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

faern:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-Ksd6wSAXgj2fGfUjikc:b-ec7gre) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

One of recent revisions changes subset to `&[&str]` slice.
</blockquote></details>

I see that. But not `haystack`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

pronebird:
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[tests/states.rs, line 19 at r3](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38#-Ksd0VNU9kz8YqDgxLbt:-KsdMl0HUzJXz2Kr_XCm:b-wnatxu) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4fdd18b5d41d781cfb3920a1a55776204d94b54/tests/states.rs#L19)):*

I see that. But not `haystack`
</blockquote></details>

Yeah hmmm, I should have changed it to slice at least. `&[String]`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/38)*

PR #40, opened 2017-08-30
Extract Transaction, ioctl_guard! and PF helpers into individual modules

pronebird:
I moved `ioctl_guard!` into `macros.rs` and functions that we use to work with PF into `utils.rs`. The reason for that is because we may incorporate PF helpers into something else in future, but `ioctl_guard!` will most likely remain standalone.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/40)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -241,7 +241,7 @@ impl PfCtl {
         let mut pfioc_rule = unsafe { mem::zeroed::<ffi::pfvar::pfioc_rule>() };
 
         pfioc_rule.pool_ticket = utils::get_pool_ticket(self.fd(), anchor)?;
-        pfioc_rule.ticket = self.get_ticket(&anchor, AnchorKind::Filter)?;
+        pfioc_rule.ticket = utils::get_ticket(self.fd(), &anchor, AnchorKind::Filter)?;
         anchor
             .try_copy_to(&mut pfioc_rule.anchor[..])
             .chain_err(|| ErrorKind::InvalidArgument("Invalid anchor name"))?;
@@ -335,17 +335,6 @@ impl PfCtl {
         let element_size = mem::size_of::<ffi::pfvar::pfsync_state>() as u32;
         let buffer_size = pfioc_states.ps_len as u32;
         Ok(buffer_size / element_size)
-    }
-
-    fn get_ticket(&self, anchor: &str, kind: AnchorKind) -> Result<u32> {
-        let mut pfioc_rule = unsafe { mem::zeroed::<ffi::pfvar::pfioc_rule>() };
-        pfioc_rule.action = ffi::pfvar::PF_CHANGE_GET_TICKET as u32;
-        pfioc_rule.rule.action = kind.into();
-        anchor
-            .try_copy_to(&mut pfioc_rule.anchor[..])
-            .chain_err(|| ErrorKind::InvalidArgument("Invalid anchor name"))?;
-        ioctl_guard!(ffi::pf_change_rule(self.fd(), &mut pfioc_rule))?;
-        Ok(pfioc_rule.ticket)
     }
 
     /// Internal function for getting the raw file descriptor to PF.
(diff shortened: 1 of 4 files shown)


faern:
:lgtm:

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/macros.rs, line 16 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40#-KssfQCuWpPB5NkQ-Str:-KssfQCvxF2uYEQI60E2:blazuvc) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0fe246fa727e29ee4acf727b9f56b4f2ffdb0f54/src/macros.rs#L16)):*
> ```Rust
>     ($func:expr, $already_active:expr) => {
>         if unsafe { $func } == $crate::macros::IOCTL_ERROR {
>             let $crate::errno::Errno(error_code) = $crate::errno::errno();
> ```

`$crate` is useful when someone else is going to executed macros in our crate, because then we need to reference back to our crate. But this macro will only execute in our own crate, and thus `::errno` should be enough, because `errno` will always be available in the root namespace inside this crate.

---

*[src/utils.rs, line 18 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40#-Kssgh8qqu2-Wr15vlEi:-Kssgh8qqu2-Wr15vlEj:b-vx1og5) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0fe246fa727e29ee4acf727b9f56b4f2ffdb0f54/src/utils.rs#L18)):*
> ```Rust
> 
> /// The path to the PF device file this library will use to communicate with PF.
> pub const PF_DEV_PATH: &'static str = "/dev/pf";
> ```

This was previously public to the entire crate, showing up in the docs etc. At the moment it's public only from this module, but not exported from the crate. So either remove the `pub` keyword here so it's contained only in here. Or if we still want to expose it then do a `pub use utils::PF_DEV_PATH` at the root.

But I don't see much reason to export this publicly, just removing the `pub` is probably enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40#-:-KsshldeIWDpxkWGMv_E:bnfp4nl)*

pronebird:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/macros.rs, line 16 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40#-KssfQCuWpPB5NkQ-Str:-KsspyRSvKPMgUx9wn0C:be8jxdo) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0fe246fa727e29ee4acf727b9f56b4f2ffdb0f54/src/macros.rs#L16)):*

`$crate` is useful when someone else is going to executed macros in our crate, because then we need to reference back to our crate. But this macro will only execute in our own crate, and thus `::errno` should be enough, because `errno` will always be available in the root namespace inside this crate.
</blockquote></details>

OK fixed that.

---

*[src/utils.rs, line 18 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40#-Kssgh8qqu2-Wr15vlEi:-KssqPjFKGeX5AzvDktI:b6oqiw4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0fe246fa727e29ee4acf727b9f56b4f2ffdb0f54/src/utils.rs#L18)):*

This was previously public to the entire crate, showing up in the docs etc. At the moment it's public only from this module, but not exported from the crate. So either remove the `pub` keyword here so it's contained only in here. Or if we still want to expose it then do a `pub use utils::PF_DEV_PATH` at the root.

But I don't see much reason to export this publicly, just removing the `pub` is probably enough.
</blockquote></details>

Simply left over. Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40)*

faern:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/40)*

PR #47, opened 2017-09-19
WIP: Add retry_if_busy helper

pronebird:
This PR adds helper that enables to repeat failed ioctl calls due to `EBUSY` error which is normally returned in the event of concurrent changes to firewall. 

Usually that may happen if one or more program modify the same anchor simultaneously. 

In that case what happens is the ticket that we previously received becomes invalid so we need to re-issue it and start over.

This is WIP since I still work on simulation tests but I wanted to share that little piece of code to gather more feedback.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -137,6 +137,43 @@ macro_rules! ignore_error_kind {
 }
 
 
+/// Delay between retries
+const RETRY_IF_BUSY_DELAY: u64 = 100;
+
+/// Maximum runtime of retry_if_busy before giving up
+const RETRY_IF_BUSY_TIMEOUT: u64 = 1000;
+
+/// Maximum number of retries to perform before giving up
+const RETRY_IF_BUSY_MAX: i8 = 5;
+
+/// Helper function that runs the given closure if received error indicates that firewall rules
+/// were modified concurrently by other program until either timeout, or number of retries reached,
+/// or any other result occurred.
+pub fn retry_if_busy<F, R>(f: F) -> Result<R>
+where
+    F: Fn() -> Result<R>,
+{
+    use std::time::{Duration, Instant};
+    let start = Instant::now();
+    let timeout = Duration::from_millis(RETRY_IF_BUSY_TIMEOUT);
+    let delay = Duration::from_millis(RETRY_IF_BUSY_DELAY);
+    let mut retry = 0;
+    loop {
+        match f() {
+            Err(Error(ErrorKind::IoctlError(ref io_err), _))
+                if io_err.raw_os_error() == Some(libc::EBUSY) &&
+                    (start.elapsed() + delay) < timeout &&
+                    retry < RETRY_IF_BUSY_MAX =>
+            {
+                retry += 1;
+                ::std::thread::sleep(delay);
+            }
+            r => return r,
+        }
+    }
+}
+
+
 /// Module for types and traits dealing with translating between Rust and FFI.
 mod conversion {
     /// Internal trait for all types that can write their value into another type without risk


faern:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuPJS4appNVKL3-icSA:b-1k5n28) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*
> ```Rust
> /// were modified concurrently by other program until either timeout, or number of retries reached,
> /// or any other result occurred.
> pub fn retry_if_busy<F, R>(f: F) -> Result<R>
> ```

How do we intend this to be used? Should a user have to wrap calls in this function manually? Would be nice if one could configure `PfCtl` and/or `Transaction` to automatically retry, so the user does not have to manually wrap everything all the time.

I'm thinking about how we want to use this. As long as there is no way `EBUSY` could be returned forever I think we should retry indefinitely on all our internal calls. Compare this with `iptables` on Linux which can also fail if multiple commands are run at the same time, but then it has a `-w` flag that will make the command block until it has exclusive access to the firewall and can complete its actions. With `-w` it never returns until it has succeeded.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuPJAXuiNq_dIjUY3Ni:bxv7ewe) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*
> ```Rust
>     F: Fn() -> Result<R>,
> {
>     use std::time::{Duration, Instant};
> ```

Why do you import here? I think we should do imports at the top of the file if there is no special reason this suits better.

---

*[src/lib.rs, line 161 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPIyqupjw3O6LxNX4s:-KuPIyqupjw3O6LxNX4t:biuqmtk) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L161)):*
> ```Rust
>     let delay = Duration::from_millis(RETRY_IF_BUSY_DELAY);
>     let mut retry = 0;
>     loop {
> ```

Can be shortened with `for _ in 0..RETRY_IF_BUSY_MAX {...}` to not have to keep a separate loop counter.

---

*[src/lib.rs, line 164 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPKUm-DHkacHGLcD9T:-KuPKUm0Aev8PA2nitRb:b71celm) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L164)):*
> ```Rust
>         match f() {
>             Err(Error(ErrorKind::IoctlError(ref io_err), _))
>                 if io_err.raw_os_error() == Some(libc::EBUSY) &&
> ```

There is no part of the PF api that can be used to grant exclusive access to PF? Like a mutex. Say you can call this mutex first and then be guaranteed to have access to the firewall?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuPLlbNrU-avZMjt_BZ:bmc4pe) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

How do we intend this to be used? Should a user have to wrap calls in this function manually? Would be nice if one could configure `PfCtl` and/or `Transaction` to automatically retry, so the user does not have to manually wrap everything all the time.

I'm thinking about how we want to use this. As long as there is no way `EBUSY` could be returned forever I think we should retry indefinitely on all our internal calls. Compare this with `iptables` on Linux which can also fail if multiple commands are run at the same time, but then it has a `-w` flag that will make the command block until it has exclusive access to the firewall and can complete its actions. With `-w` it never returns until it has succeeded.
</blockquote></details>

Roughly as:

```
let result = retry_if_busy(|| {
  let t = Transaction::new();
  t.add_change(...);
  t.commit()
});
```

Certainly on internal calls. I am a little afraid to create infinite loops on `EBUSY`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuPMMKul8UB00-Tpk9G:bqzvfk9) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

Why do you import here? I think we should do imports at the top of the file if there is no special reason this suits better.
</blockquote></details>

Because it's not used anywhere outside of this function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 161 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPIyqupjw3O6LxNX4s:-KuPMPvQxELpWPl1qIV9:b-kipptb) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L161)):*

Can be shortened with `for _ in 0..RETRY_IF_BUSY_MAX {...}` to not have to keep a separate loop counter.
</blockquote></details>

Not without explicitly returning `Result` from the enclosing function. `loop` on the other side removes that requirement, we rely on `return` statement instead of natural flow of finite loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/lib.rs, line 164 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPKUm-DHkacHGLcD9T:-KuPNJMu7LfB2Ps5ropw:b-xhc7kh) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L164)):*

There is no part of the PF api that can be used to grant exclusive access to PF? Like a mutex. Say you can call this mutex first and then be guaranteed to have access to the firewall?
</blockquote></details>

Cross-process mutex? That would be incredibly dangerous. The only mutex we get is the ticket, if ticket becomes invalid then firewall returns `EBUSY`. So we have to get new tickets and repeat the dance. :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

faern:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuPRPsGu4YbDSsl2jz1:bmolte6) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

Roughly as:

```
let result = retry_if_busy(|| {
  let t = Transaction::new();
  t.add_change(...);
  t.commit()
});
```

Certainly on internal calls. I am a little afraid to create infinite loops on `EBUSY`.
</blockquote></details>

But most people will want their operations to go through. And it would be extremely cumbersome to have to manually do what you posted now in every place a user uses PF.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuPQWVTZ-xizjSUM0fj:bzejver) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

Because it's not used anywhere outside of this function.
</blockquote></details>

But this is still a new way to do it? We have always imported at the top of the file in all other crates. What made you do it differently here? We can keep it if you like it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

faern:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuPRlLSB7DWhy-gGqId:b-9ijwr1) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

But most people will want their operations to go through. And it would be extremely cumbersome to have to manually do what you posted now in every place a user uses PF.
</blockquote></details>

Plus, they would need to re-create all their rules in the part that is `t.add_change(...)`. This could introduce complexity for the end user and will cost a lot of allocation and processing. If `t.commit()` could do it internally it could just reuse all the already created rules for example. Just re-create the ffi versions of them or similar.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuQM0Dd0hfKQ-WxgodE:b-30oagf) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

Plus, they would need to re-create all their rules in the part that is `t.add_change(...)`. This could introduce complexity for the end user and will cost a lot of allocation and processing. If `t.commit()` could do it internally it could just reuse all the already created rules for example. Just re-create the ffi versions of them or similar.
</blockquote></details>

I don't think it's only about transactions. It's relevant to any PF call that changes firewall rules. 

Yes it's complexity on the end user because it's up to the end user what to do. 

For example end user may decide to reevaluate the firewall fules before proceeding in case of changes happened externally.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuQMYYiMMnCW9n81FVo:b-1aknl8) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

I don't think it's only about transactions. It's relevant to any PF call that changes firewall rules. 

Yes it's complexity on the end user because it's up to the end user what to do. 

For example end user may decide to reevaluate the firewall fules before proceeding in case of changes happened externally.
</blockquote></details>

Transactions are actually great example of complex operation breaking in the middle due to concurrent changes for PF since transaction issues multiple `pf_add_rule` calls with the obtained tickets. 

Half-way through it may simply break because other process was faster. So one would have to start over, get tickets once again, re-create all FFI structs etc..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuQNRtbADdxt0UUUjGk:bpt665d) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

But this is still a new way to do it? We have always imported at the top of the file in all other crates. What made you do it differently here? We can keep it if you like it.
</blockquote></details>

I don't know, I didn't want to pollute the namespace but I have no problem to fair to move it up.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuQNXd3ZwN2TIOFzzL1:b-lc21yd) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

I don't know, I didn't want to pollute the namespace but I have no problem to fair to move it up.
</blockquote></details>

to fair = to be fair

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuQNh5f7a0iErm9iEUL:b7rbh24) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

to fair = to be fair
</blockquote></details>

Also, it was macro in the beginning :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 156 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJAXtNtx_ySUVmmWW:-KuQO-QdXllf4ae1DYHz:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L156)):*

Also, it was macro in the beginning :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuQOPJJM0kaOkdN70VB:blpulls) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

Transactions are actually great example of complex operation breaking in the middle due to concurrent changes for PF since transaction issues multiple `pf_add_rule` calls with the obtained tickets. 

Half-way through it may simply break because other process was faster. So one would have to start over, get tickets once again, re-create all FFI structs etc..
</blockquote></details>

I may be overthinking this a little, but I thought that magic loops brute forcing the firewall was something that user would rather do explicitly...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 141 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuQOvGpWaF9XkvBdlau:-KuQOvGpWaF9XkvBdlav:b3rrzhk) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L141)):*
> ```Rust
> 
> /// Delay between retries
> const RETRY_IF_BUSY_DELAY: u64 = 100;
> ```

When we first discussed this issue we looked towards number of retries limitation and timeout. 

However I think we can actually remove the timeout because it's irrelevant, it's not network request and local calls shouldn't take longer than a split second to be processed by computer. Limiting this to number of attempts should be sensible enough IMO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

pronebird:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/lib.rs, line 141 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuQOvGpWaF9XkvBdlau:-KuQPN_ddampOLPktpGn:b-ys0rgv) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L141)):*

When we first discussed this issue we looked towards number of retries limitation and timeout. 

However I think we can actually remove the timeout because it's irrelevant, it's not network request and local calls shouldn't take longer than a split second to be processed by computer. Limiting this to number of attempts should be sensible enough IMO.
</blockquote></details>

We can still keep a short delay to let firewall process changes coming from all processes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

faern:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 152 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuPJS4appNVKL3-icS9:-KuSrMIqNMIQsMGbU7hl:b-gnwoq4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L152)):*

I may be overthinking this a little, but I thought that magic loops brute forcing the firewall was something that user would rather do explicitly...
</blockquote></details>

One thing that is important to know to evaluate this in a fair way: Can `EBUSY` happen if someone else edit any part of the firewall, or does it only happen if the two processes try to modify the exact same rules/anchor?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

faern:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/lib.rs, line 141 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47#-KuQOvGpWaF9XkvBdlau:-KuUMA7FtU1IZ6Oo3Wlf:b-wk5y2l) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c74f42c64c070464ae9e101af305cbb566ad8068/src/lib.rs#L141)):*

We can still keep a short delay to let firewall process changes coming from all processes.
</blockquote></details>

I agree. Timer is not really needed IMO. At least according to my understanding of when/how `EBUSY` happens.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/47)*

PR #48, opened 2017-09-20
Redirect interface

faern:
Adding support for the log and the interface arguments to redirect rules.

Also made interface names its own type, according to the newtype pattern. No reason for this really, I thought it would be needed for `route-to` support later, but not sure yet. Still just as ergonomic to use `Interface` because of the nifty impls.


pronebird:
Reviewed 1 of 5 files at r1.
Review status: 1 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[.travis.yml, line 20 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUan8rawUDHvQdVDdc:-KuUan8rawUDHvQdVDdd:b-7hwzak) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/.travis.yml#L20)):*
> ```YAML
>   # Format only on nightly, since that is where rustfmt-nightly compiles
>   - if [ "${TRAVIS_RUST_VERSION}" = "nightly" ]; then
>       ./format.sh --write-mode=diff;
> ```

Does it work without exporting`DYLD_LIBRARY_PATH`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Reviewed 1 of 5 files at r1.
Review status: 2 of 5 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Review status: 2 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/rule/interface.rs, line 13 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUb8mNldrSYA7zEUq2:-KuUb8mNldrSYA7zEUq3:bndnaqq) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/src/rule/interface.rs#L13)):*
> ```Rust
> 
> #[derive(Debug, Clone, PartialEq, Eq, Hash)]
> pub struct InterfaceName(String);
> ```

What's the practical use of that? Is it the same as using `AsRef<String>`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Review status: 2 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/redirect_rules.rs, line 95 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUbpbRCRQzwkUnAag0:-KuUbpbRCRQzwkUnAag1:bhhpqrz) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/tests/redirect_rules.rs#L95)):*
> ```Rust
>     assert_matches!(
>         pfcli::get_nat_rules(ANCHOR_NAME),
>         Ok(ref v) if v == &["rdr log on lo0 all -> any port 1237"]
> ```

Not that long ago you asked me to have less dangerous rules in tests. Would you mind adding some condition to that rule to ensure that we don't turn all traffic on `lo0` to single port.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Reviewed 3 of 5 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

faern:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[.travis.yml, line 20 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUan8rawUDHvQdVDdc:-KuUcL3e0ntDUO3I6ynk:b-7n3fyy) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/.travis.yml#L20)):*

Does it work without exporting`DYLD_LIBRARY_PATH`?
</blockquote></details>

This change should not be here. It's part of the other PR. Just waiting for travis to finish. Please ignore this file in this PR.

---

*[src/rule/interface.rs, line 13 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUb8mNldrSYA7zEUq2:-KuUcUirlU_aXsF2wBPW:bx60zx4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/src/rule/interface.rs#L13)):*

What's the practical use of that? Is it the same as using `AsRef<String>`?
</blockquote></details>

An `Interface` is an optional interface. I thought we would need a non-optional interface in the `route-to` impl, so I created this. But I'm not sure we will need it.

If we need it it's better to express it in the type system than just using `String` everywhere. It expresses the semantic meaning. See https://doc.rust-lang.org/book/first-edition/structs.html#tuple-structs But if we never need `InterfaceName` explicitly we can ignore creating it. Let's see what the research on `route-to` gives us.

---

*[tests/redirect_rules.rs, line 95 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUbpbRCRQzwkUnAag0:-KuUd6kRKsYvbOh2g9HQ:bilrah8) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/tests/redirect_rules.rs#L95)):*

Not that long ago you asked me to have less dangerous rules in tests. Would you mind adding some condition to that rule to ensure that we don't turn all traffic on `lo0` to single port.
</blockquote></details>

Yes. My bad. Will do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/rule/interface.rs, line 13 at r1](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-KuUb8mNldrSYA7zEUq2:-KuUdn2DjvVCxcBU98Lk:b9si2ge) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f1f036002c1a051ae9e540285459fca09eb52261/src/rule/interface.rs#L13)):*

An `Interface` is an optional interface. I thought we would need a non-optional interface in the `route-to` impl, so I created this. But I'm not sure we will need it.

If we need it it's better to express it in the type system than just using `String` everywhere. It expresses the semantic meaning. See https://doc.rust-lang.org/book/first-edition/structs.html#tuple-structs But if we never need `InterfaceName` explicitly we can ignore creating it. Let's see what the research on `route-to` gives us.
</blockquote></details>

hmmmm OK makes sense to have it as marker type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

pronebird:
Once tests are fixed it becomes :point_right:  :lgtm: 

---

Review status: 4 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48#-:-KuUe6dgI6DwG11Yf6po:b-6cuiwr)*

pronebird:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/48)*

PR #49, opened 2017-09-20
WIP: Add retry_if_busy helper (take 2)

pronebird:
This PR adds helper that enables to repeat failed ioctl calls due to EBUSY error which is normally returned in the event of concurrent changes to firewall.

Usually that may happen if one or more program modify the same anchor simultaneously.

In that case what happens is the ticket that we previously received becomes invalid so we need to re-issue it and start over.

This is WIP since I still work on simulation tests but I wanted to share that little piece of code to gather more feedback.

Original reviewable board: https://reviewable.io/reviews/mullvad/pfctl-rs/47

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/49)

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -20,6 +20,7 @@ use utils;
 #[derive(Debug)]
 pub struct Transaction {
     change_by_anchor: HashMap<String, AnchorChange>,
+    retry_on_concurrent_changes: bool,
 }
 
 impl Transaction {
@@ -27,7 +28,14 @@ impl Transaction {
     pub fn new() -> Self {
         Transaction {
             change_by_anchor: HashMap::new(),
+            retry_on_concurrent_changes: false,
         }
+    }
+
+    /// Allows transaction to commit more than once on failure due to concurrent changes to
+    /// affected anchors. Defaults to false.
+    pub fn set_retry_on_concurrent_changes(&mut self, retry: bool) {
+        self.retry_on_concurrent_changes = retry;
     }
 
     /// Add change into transaction replacing the prior change registered for corresponding
@@ -41,7 +49,6 @@ impl Transaction {
     pub fn commit(mut self) -> Result<()> {
         let pf_file = utils::open_pf()?;
         let fd = pf_file.as_raw_fd();
-        let mut pfioc_trans = unsafe { mem::zeroed::<ffi::pfvar::pfioc_trans>() };
 
         // partition changes by ruleset kind
         let filter_changes: Vec<(String, Vec<FilterRule>)> = self.change_by_anchor
@@ -62,6 +69,21 @@ impl Transaction {
                     .map(|rules| (anchor.clone(), rules))
             })
             .collect();
+
+        if self.retry_on_concurrent_changes {
+            retry_on_busy!(Self::try_commit(fd, &filter_changes, &redirect_changes))
+        } else {
+            Self::try_commit(fd, &filter_changes, &redirect_changes)
+        }
+    }
+
+    /// Internal helper that converts series of changes into FFI transaction and commits it.
+    fn try_commit(
+        fd: RawFd,
+        filter_changes: &Vec<(String, Vec<FilterRule>)>,
+        redirect_changes: &Vec<(String, Vec<RedirectRule>)>,
+    ) -> Result<()> {
+        let mut pfioc_trans = unsafe { mem::zeroed::<ffi::pfvar::pfioc_trans>() };
 
         // create one transaction element for each unique combination of anchor name and
         // `RulesetKind` and order them so elements for filter rules go first followed by redirect
@@ -84,7 +106,7 @@ impl Transaction {
         let mut ticket_iterator = pfioc_elements.iter().map(|e| e.ticket);
 
         // add filter rules into transaction
-        for ((anchor_name, filter_rules), ticket) in
+        for (&(ref anchor_name, ref filter_rules), ticket) in
             filter_changes.into_iter().zip(ticket_iterator.by_ref())
         {
             for filter_rule in filter_rules.iter() {
@@ -93,7 +115,7 @@ impl Transaction {
         }
 
         // add redirect rules into transaction
-        for ((anchor_name, redirect_rules), ticket) in
+        for (&(ref anchor_name, ref redirect_rules), ticket) in
             redirect_changes.into_iter().zip(ticket_iterator.by_ref())
         {
             for redirect_rule in redirect_rules.iter() {


pronebird:
I thought about the same thing regarding retrying requests! Same can be added for `PfCtl` too. I hope I don't swim against the current here because I really think that it's up to library consumer to decide what to do in case if some other process changed firewall rules at the same time.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-:-KusRFPv60mCG62rhXg9:bu4rxna)*

pronebird:
@faern too verbose? `set_retry_on_concurrent_changes(&self, retry: bool)`

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-:-KusfhNeQIjPGODB_Nfp:b-1mflbs)*

pronebird:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/transaction.rs, line 109 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KusiHc9Y5vVk36BLuWG:-KusiHc9Y5vVk36BLuWH:bnra3ua) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/transaction.rs#L109)):*
> ```Rust
> 
>         // add filter rules into transaction
>         for (&(ref anchor_name, ref filter_rules), ticket) in
> ```

Temporarily switched to reference to avoid `move`. Maybe I can improve that and get rid of refs while keeping this helper function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49)*

pronebird:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/transaction.rs, line 74 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KusiQ-YAohFjGiJMiYS:-KusiQ-YAohFjGiJMiYT:b-q3cgsb) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/transaction.rs#L74)):*
> ```Rust
> 
>         if self.retry_on_concurrent_changes {
>             retry_on_busy!(Self::try_commit(fd, &filter_changes, &redirect_changes))
> ```

`retry_on_busy` macro comes real handy here!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49)*

pronebird:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/transaction.rs, line 83 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KustVEGUJN8rAh4xfpo:-KustVEGUJN8rAh4xfpp:bdu3wvg) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/transaction.rs#L83)):*
> ```Rust
>     fn try_commit(
>         fd: RawFd,
>         filter_changes: &Vec<(String, Vec<FilterRule>)>,
> ```

Linus if you have suggestion how to get rid of this, I am all ears :D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49)*

faern:
Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/macros.rs, line 30 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KvxAE2ZAae4Dqw6nvSQ:-KvxAE2ZAae4Dqw6nvSR:bmm1c8p) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/macros.rs#L30)):*
> ```Rust
> 
> /// Delay between retries (in milliseconds)
> pub const RETRY_IF_BUSY_DELAY: u64 = 100;
> ```

Do we need a delay? Is this something you have read or something you have thought about yourself? Is it not that a transaction becomes invalid only if someone else **commits** during your transaction? Thus the other operation that renders your invalid should be atomic, right? If so retrying instantly would be fine, because we know the other one is already done.

---

*[src/macros.rs, line 33 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KvxAeasHiKDhgc8QCIK:-KvxAeasHiKDhgc8QCIL:bs40gjb) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/macros.rs#L33)):*
> ```Rust
> 
> /// Maximum number of retries to perform before giving up
> pub const RETRY_IF_BUSY_MAX: i8 = 5;
> ```

I think retries should be infinite. Otherwise we still have the `EBUSY` error case to deal with as a consumer of this library. If I set the setting to retry until it works I expect it to try until it works. Do we have any reason to suspect retrying infinitely could actually get stuck indefinitely or be bad in some other way?

This was partly the reason why I suggested an enum instead of bool, because then we can have:
```rust
pub enum RetrySetting {
    NoRetry,
    Retry(u32),
    RetryIndefinitely,
}
```
To let the consumers decide if it should be a limited number of times or until it works.

---

*[src/transaction.rs, line 83 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KustVEGUJN8rAh4xfpo:-KvxC2YxlSt8cEwh4Htn:b-wto4zk) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/transaction.rs#L83)):*

Linus if you have suggestion how to get rid of this, I am all ears :D
</blockquote></details>

Get rid of what? This helper function?

Btw: Use `&[]` instead of `&Vec<>` if it's just borrowed data.

---

*[src/transaction.rs, line 109 at r2](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49#-KusiHc9Y5vVk36BLuWG:-KvxCNCr8Jv0dQCWxZFy:btwjkxv) ([raw file](https://github.com/mullvad/pfctl-rs/blob/71161e55e7311f93d61b16ed7acb30354244201c/src/transaction.rs#L109)):*

Temporarily switched to reference to avoid `move`. Maybe I can improve that and get rid of refs while keeping this helper function.
</blockquote></details>

I think the problem is that you try to do `into_iter` when you have borrowed data. If you change the argument to the function to a slice instead of a vector and then use `.iter` I think you can get rid of the `ref`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/49)*

pronebird:
@faern I believe we should close this PR.

PR #58, opened 2018-01-05
Commit generated pfvar

faern:
Generating the the bindings with bindgen on build had the problem that it was dependent on finding the macOS SDK path. Thus the library could only be build on a machine with xcode installed. This is unfortunate. And it blocks docs.rs from building the docs as well: https://docs.rs/pfctl/0.1.0/pfctl/

bindgen recommends building bindings on the fly with a buildscript, like we do before this PR: https://rust-lang-nursery.github.io/rust-bindgen/tutorial-1.html, but that does not work very well for us. Also, these bindings are quite unlikely to change. Or at least the C source header is. The generated code can change because of changes to bindgen or rustfmt of course.

What I did was to change the code in `build.rs` into a bash script calling the bindgen binary instead to generate the same binding result, but commit that resulting binding as well. So if we have `pfvar.h` and `generate_bindings.sh` in the repo it becomes very easy to re-generate the bindings on bindgen updates or similar. But for ordinary people who just want to build the library they are now not required to have xcode and they also don't have to build bindgen at all.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/58)

PR #59, opened 2018-01-05
Dont depend on xcodebuild

faern:
So I found this less dramatic change to get the same result as #58. However, I'm not determined as to which solution is the best one yet.

You know more about macOS than me, will this solution work on a machine without xcode or the SDK installed? Will the global `/System/Library/Frameworks/Kernel.framework/Versions/A/Headers` folder exist on such machine, or is it just mine+travis because we also have xcode installed already?

This solution is cleaner in the sense that we don't have to commit 4000 lines of generated Rust code. One con of this solution is that the crate can still only be built on a macOS machine. If we did not depend on the headers at all and checked in the generated code it would likely build (but not run) on any platform, which would make it a bit easier to depend on this crate and to generate the docs on non-MACs.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/59)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,23 +2,17 @@ extern crate bindgen;
 
 use std::env;
 use std::path::Path;
-use std::process::Command;
-use std::str;
 
+#[cfg(target_os = "macos")]
 fn main() {
     let out_dir = env::var("OUT_DIR").expect("OUT_DIR missing from environment");
     println!("OUT_DIR: {}", out_dir);
-    let sdk_path = get_macos_sdk_path();
-    println!("sdk_path: {}", sdk_path);
 
     let _ = bindgen::builder()
         .header("ffi/pfvar.h")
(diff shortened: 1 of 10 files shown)


pronebird:
I think you need command line tools on macOS for `/System/Library/Frameworks/Kernel.framework/Headers` to be installed.

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/59#-:-L2Kitn4ZxyHCOGveaAG:be4bb71)*

pronebird:
Reviewed 10 of 10 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/mullvad/pfctl-rs/59)*

faern:
Closed in favor of the solution in #58

PR #80, opened 2021-10-08
Fix: Don't set the `code` field for ICMP types that don't use it

faern:
Fixes a bug in #79 where we always set the `pf_rule::code` field to a non-zero value. The value zero indicates that the rule should not even check the code field. For most ICMP `type` values there is no corresponding `code` value and it does not make much sense to check it.

Both this and the previous design prevents rules from checking the code value for arbitrary numbers. A theoretical rule could be `icmp-type echoreq code 123`. It would be a valid rule and there could theoretically be packets matching it, but it would not make much sense from a practical standpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/80)

### DIFF
--- a/examples/add_rules.rs
+++ b/examples/add_rules.rs
@@ -48,6 +48,15 @@ fn run() -> Result<()> {
         .icmp_type(pfctl::IcmpType::EchoReq)
         .build()
         .unwrap();
+    let pass_all_icmp_port_unreach = FilterRuleBuilder::default()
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/80)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #88, opened 2024-04-30
Change `error-chain` to `thiserror`

### LINKED ISSUE #66: Remove dependency on `error-chain`
Rewrite the errors to not use the deprecated and defunct `error-chain` library.

RoDmitry:
Changed `error-chain` to `thiserror`, and updated dependencies.
Maybe you would also want to remove `error-chain` from examples and tests.

Closes #66

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/88)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,13 +16,14 @@ travis-ci = { repository = "mullvad/pfctl-rs" }
 
 [dependencies]
 errno = "0.2"
-error-chain = "0.12"
 ioctl-sys = "0.6.0"
 libc = "0.2.29"
 derive_builder = "0.9"
 ipnetwork = "0.16"
+thiserror = "1"
 
 [dev-dependencies]
 assert_matches = "1.1.0"
+error-chain = "0.12"
 uuid = { version = "0.8", features = ["v4"] }
 scopeguard = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,15 +59,14 @@
 
 #![deny(rust_2018_idioms)]
 
-#[macro_use]
-pub extern crate error_chain;
-
 use std::{
     ffi::CStr,
+    fmt::{Debug, Display},
     fs::File,
     mem,
     os::unix::io::{AsRawFd, RawFd},
 };
+use thiserror::Error;
 
 pub use ipnetwork;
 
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -13,16 +13,16 @@ edition = "2018"
...
-derive_builder = "0.9"
-ipnetwork = "0.16"
+thiserror = "1"
 
 [dev-dependencies]
 assert_matches = "1.1.0"
-uuid = { version = "0.8", features = ["v4"] }
+error-chain = "0.12"
 scopeguard = "1.0"
+uuid = { version = "1", features = ["v4"] }
```
faern, line 28: Please separate the dependency upgrades not related to swapping out `error-chain` to a separate PR. I prefer if PRs are as focused as possible on one thing.

src/lib.rs
```diff
@@ -92,40 +91,87 @@ pub use crate::ruleset::*;
...
+    #[error("Ioctl Error")]
+    Ioctl(#[from] ::std::io::Error),
+}
+
+#[derive(Error, Debug)]
+pub struct Error {
+    pub info: Option<ErrorInfo>,
+    #[source]
+    pub source: ErrorSource,
+}
```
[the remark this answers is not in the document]
RoDmitry: `Do you have references to other libraries using this style?`
No. I guess I've invented it myself. Just saved the original logic as it was before.

Cargo.toml
```diff
@@ -13,16 +13,16 @@ edition = "2018"
...
 ipnetwork = "0.16"
+libc = "0.2.29"
+thiserror = "1"
 
 [dev-dependencies]
 assert_matches = "1.1.0"
-uuid = { version = "0.8", features = ["v4"] }
+error-chain = "0.12"
 scopeguard = "1.0"
+uuid = { version = "0.8", features = ["v4"] }
```
faern, line 28: You still move stuff around a lot here. So there are still code changes even if the resulting dependencies are the same. I would appreciate if unrelated code is not changed :pray:

src/lib.rs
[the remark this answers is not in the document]
faern: What's the idea behind having `info` as a separate thing here? Absolutely everything regarding the internal details of these errors are exposed, making it a breaking change to change anything in it. I'd advise you to take a look at the PR I linked above where we did error improvements.

There is no need to keep the logic similar to how the errors were before. That is really old and outdated code, created by a very outdated error library. So it's probably not a good source of inspiration.

src/lib.rs
```diff
@@ -92,40 +91,89 @@ pub use crate::ruleset::*;
...
-            IoctlError(::std::io::Error);
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if let Some(info) = self.info.as_ref() {
+            info.fmt(f)?;
         }
+        Ok(())
```
faern, line 146: This means that an error with no `info` prints as nothing (empty string). That's not good. An error should never have an empty display representation, there must always be some description of what the error means.

faern (changes requested):

RoDmitry:
Fixed?

RoDmitry:
> Absolutely everything regarding the internal details of these errors are exposed

Still don't understand why is it bad

Cargo.toml
RoDmitry: It was a simple `cargo sort`

Cargo.toml
RoDmitry: I'll have to wait for the merge then. That's actually very annoying if you accept only one thing per PR.

faern: If we have issues/complications it's going to be significantly easier to find what the cause is if changes are separated into smaller logical chunks. Also easier to revert single things instead of a blob of unrelated things. Also way easier to review and reason about.

faern: If you submit small easy to review PRs I can review and merge them fast.

faern:
> > Absolutely everything regarding the internal details of these errors are exposed
> 
> Still don't understand why it's bad

Again, see https://github.com/mullvad/udp-over-tcp/pull/57. If every aspect of the error is part of the public API then every small "internal" change becomes a breaking change. Let's say we wanted to change `InvalidRuleCombination(String),` into `InvalidRuleCombination(&'static str),` for *some reason*. Because it's part of the public API we now need to release this as a breaking change (`0.5.0` instead of `0.4.7`). Because otherwise it would break compilation for users who might hypothetically destructure the error and rely on the embedded enum variant type to be a `String`.

If we don't expose details that a user usually won't ever need to rely on, the API surface both becomes cleaner, and we are freer to change the internals without breaking stuff.

RoDmitry:
I see the reasoning, but I don't understand why would you lose a part of the error. For `InvalidRuleCombination(String)`, string is the part of the error's display implementation. If you drop it, how would you display it? And also I don't understand why would anybody match and rely on the internal string, if they don't need it. Maybe they actually would need that string? Because that string makes this variant different from the same variant with the different string. It's not as "internal" as you claim.

RoDmitry:
If that is "internal", but internally we don't use it, maybe we actually don't need it then?

src/rule/mod.rs
```diff
@@ -128,14 +128,16 @@ impl FilterRule {
                     "StatePolicy {:?} and protocol {:?} are incompatible",
                     state_policy, proto
                 );
-                bail!(ErrorKind::InvalidRuleCombination(msg));
+                Err(Error::InvalidRuleCombination(msg))
```
pronebird, line 131: Would it be better to return `Error::InvalidRuleCombination(StatePolicy, Proto)` as opposed to a pre-reformatted string msg?

src/lib.rs
```diff
@@ -92,40 +91,117 @@ pub use crate::ruleset::*;
...
+    #[error("Incompatible values in rule: {}", _0)]
+    InvalidRuleCombination(String),
+
+    #[error("Anchor does not exist")]
+    AnchorDoesNotExist,
+
+    #[error("Cstr not null terminated")]
+    CstrNotTerminated,
+
+    #[error("Anchor does not exist")]
```
pronebird, line 112: Type conversion error

src/lib.rs
```diff
@@ -92,40 +91,117 @@ pub use crate::ruleset::*;
...
+}
+
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum TryCopyToError {
+    #[error("Lower port is greater than upper port.")]
+    InvalidPortRange,
+
+    #[error("String does not fit destination")]
+    StrCopyNotFits,
```
pronebird, line 126: I'd rename it to `InsufficientStringBufferLength`

src/lib.rs
```diff
@@ -92,40 +91,117 @@ pub use crate::ruleset::*;
...
+
+    #[error("String has null byte")]
+    StrCopyNullByte,
```
pronebird, line 129: I'd rename it to `StringContainsNullByte` and describe as "String should not contain null byte"

src/lib.rs
RoDmitry: fixed

src/lib.rs
RoDmitry: fixed

src/lib.rs
RoDmitry: fixed

src/rule/mod.rs
RoDmitry: I don't think the end user needs AddrFamily, because it can be incompatible only when they don't match.
I can create a separate variant for `validate_state_policy` with StatePolicy and Proto, if @faern would approve this. And also separate variants for FilterRuleBuilder and RedirectRuleBuilder.

faern:
Thanks for wanting to contribute! However, we merged a different solution to this issue over in #107. The main issue in this PR was how the entire API of the error was still public. Part of wanting to get rid of `error-chain` was that we wanted the API of the errors ta change. A much less public API surface of the errors allows changing the error more without breaking the API.

PR #89, opened 2024-05-03
Add github actions CI and remove travis CI

faern:
The CI for this crate has not been running for a looong time. We have not used Travis for years.

This PR:
* Removes the defunct Travis CI
* Adds github CI that checks for:
  * Rust formatting
  * Audit dependencies
  * Build and run tests
  * Builds with minimal dependencies
  * Git commit message style
* Checks in the lockfile for a more consistent CI experience

I wish we could also run `cargo clippy`. Hopefully we can in the future. But currently clippy complains about so much that I leave it as a separate exercise.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/89)

PR #96, opened 2024-05-29
Build in CI with --locked

faern:
Small follow-up to #89 

This ensures the Cargo.lock file stays up to date. If someone add or change a dependency in a way that require lockfile changes, the CI should fail. When the lockfile is versioned it has to stay up to date

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,7 +29,7 @@ jobs:
           default: true
 
       - name: Build
-        run: cargo build --all-targets
+        run: cargo build --all-targets --locked
 
       - name: Test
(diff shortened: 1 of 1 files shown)


Serock3:
I suppose we don't want to force gpg key signatures on commits that change the lockfile, right?

faern:
> I suppose we don't want to force gpg key signatures on commits that change the lockfile, right?

We don't have the infrastructure in place to check GPG signatures anywhere but our main repository. The keys are locally checked in there and that's why we can do it there. We have not worked on generalizing it to be able to check other repositories.

We don't really check in the lockfile in this crate for supply chain security. More for CI and debug stability/reproducibility.

PR #98, opened 2024-05-29
Replace errno dependency with calling Error::last_os_error

faern:
The documentation for the errno crate specify that `last_os_error` fetch the underlying errno in the exact same way, so this should be equivalent, but with fewer dependencies. See here: https://github.com/lambda-fairy/rust-errno?tab=readme-ov-file#comparison-with-stdioerror

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/98)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "rustc-demangle",
- "winapi 0.3.6",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Serock3 (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/98)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #102, opened 2024-05-29
Stop using error_chain! in examples

faern:
We want to get rid of `error-chain` as a dependency completely. Some work towards this has been started in #88. However, for now that PR keeps `error-chain` for the examples. I approach the problem from the other direction with this PR: Getting rid of `error-chain` in the examples first. They are simpler and require less nice API surface since they are just example binaries, not a public library API.

Here I opted for simply going with `Result<..., Box<dyn std::error::Error>>`. This has the positive side of being very simple. But with the downside that we don't have an error *chain* in the same way. We lose some of the context we previously had with an error. This might make it slightly harder to track down what failed when examples fail. On the other hand, examples should focus on showcasing how to use the main crate, not how to do ideal error handling. I'm open to discuss alternative solutions. We could replace all `chain_err` with `expect` and have the examples panic instead. In practice it won't make much of a difference since that's basically what returning an error from main does anyway :shrug:. However, that's arguably even more "ugly" error handling than `Box<dyn Error>`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102)

### DIFF
--- a/examples/add_anchor.rs
+++ b/examples/add_anchor.rs
@@ -6,25 +6,18 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#[macro_use]
-extern crate error_chain;
-
 use pfctl::PfCtl;
 use std::env;
 
-error_chain! {}
-quick_main!(run);
-
-fn run() -> Result<()> {
-    let mut pf = PfCtl::new().chain_err(|| "Unable to connect to PF")?;
+fn main() {
+    let mut pf = PfCtl::new().expect("Unable to connect to PF");
 
     for anchor_name in env::args().skip(1) {
         pf.try_add_anchor(&anchor_name, pfctl::AnchorKind::Filter)
-            .chain_err(|| "Unable to add filter anchor")?;
+            .expect("Unable to add filter anchor");
         pf.try_add_anchor(&anchor_name, pfctl::AnchorKind::Redirect)
-            .chain_err(|| "Unable to add redirect anchor")?;
+            .expect("Unable to add redirect anchor");
 
         println!("Added {} as both a redirect and filter anchor", anchor_name);
     }
-    Ok(())
 }
--- a/examples/add_rules.rs
+++ b/examples/add_rules.rs
@@ -6,23 +6,17 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-#[macro_use]
-extern crate error_chain;
-
 use pfctl::{ipnetwork, FilterRuleBuilder, PfCtl, RedirectRuleBuilder};
 use std::net::Ipv4Addr;
 
-error_chain! {}
-quick_main!(run);
-
 static ANCHOR_NAME: &str = "test.anchor";
 
-fn run() -> Result<()> {
-    let mut pf = PfCtl::new().chain_err(|| "Unable to connect to PF")?;
+fn main() {
+    let mut pf = PfCtl::new().expect("Unable to connect to PF");
     pf.try_add_anchor(ANCHOR_NAME, pfctl::AnchorKind::Filter)
-        .chain_err(|| "Unable to add test filter anchor")?;
+        .expect("Unable to add test filter anchor");
     pf.try_add_anchor(ANCHOR_NAME, pfctl::AnchorKind::Redirect)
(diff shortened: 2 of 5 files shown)


Serock3 (approved):
Personally, I prefer `expect`/`unwrap` over `Box<dyn std::error::Error>` in tests because of the verbose error formatting. It's a matter of taste so I won't block the PR on this

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-:-NzTSJTx6Wri_EpYX3bD:b-nlqrei)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
Do you mean like this? I don't regard examples as "tests". They are supposed to showcase how to use the library. And this way it's littered with unidiomatic error handling. But yes, I agree that it both produce better debug info on failure, and it contains no boilerplate just for doing sane errors in the examples, so it's probably the least bad solution :+1:

> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-:-O-6e7MkEl_71I60cjOi:bte6arg)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @Serock3)

Serock3 (approved):
Sure, an argument could be made for showcasing idiomatic error handling in examples, but in my opinion `?` doesn't really accomplish that on its own, and it would only be necessary if the error handling is non-trivial.

It's a shame that the `expect`s often break the line and make it slightly more verbose. I think being even more unidiomatic with`unwrap`s instead could be worth it for readability, but I'll leave it up to you to decide.

> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-:-O-AIBaq-snsWbJhny8n:b-3f2hu5)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`examples/flush_rules.rs` line 16 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-O-AGcFCEcgD3BE1PZ9v:-O-AGcFCEcgD3BE1PZ9w:bo9ed0n) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f5c3042ce2e277b3f053ee3d37f04bfce785f54f/examples/flush_rules.rs#L16)):*
> ```XML
> 
>     for anchor_name in env::args().skip(1) {
>         match pf.flush_rules(&anchor_name, pfctl::RulesetKind::Filter) {
> ```

What is the purpose of this match statement? Is it not equivalent to 
```rust
pf.flush_rules(&anchor_name, pfctl::RulesetKind::Filter).expect("Unable to flush filter rules");
println!("Flushed filter rules under anchor {}", anchor_name)
```

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`examples/flush_rules.rs` line 16 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-O-AGcFCEcgD3BE1PZ9v:-O-AmAcV0HGSmmgUsX_M:bcp143y) ([raw file](https://github.com/mullvad/pfctl-rs/blob/f5c3042ce2e277b3f053ee3d37f04bfce785f54f/examples/flush_rules.rs#L16)):*
<details><summary><i>Previously, Serock3 (Sebastian Holmin) wrote…</i></summary><blockquote>

What is the purpose of this match statement? Is it not equivalent to 
```rust
pf.flush_rules(&anchor_name, pfctl::RulesetKind::Filter).expect("Unable to flush filter rules");
println!("Flushed filter rules under anchor {}", anchor_name)
```
</blockquote></details>

Yeah, that makes sense. I can't remember exactly why this was written this way. Fixed.

Serock3 (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/102#-:-O-BKl-r1KA--PgYJ99j:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #104, opened 2024-06-27
Remove error-chain dependency in tests and test helpers

faern:
Follow up to #102. I realized we also used `error-chain` in tests :facepalm:. So a step in the direction of removing error-chain completely is to stop using it in tests. I just replaced most error handling with `expect`, since expect/unwrap is all we did with the errors we had anyway.

Since a lot of the test helper methods became infallible, sadly virtually every test was affected, due to them using the test helpers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/104)

### DIFF
--- a/tests/anchors.rs
+++ b/tests/anchors.rs
@@ -1,7 +1,4 @@
 #[macro_use]
-extern crate error_chain;
-
-#[macro_use]
 #[allow(dead_code)]
 mod helper;
 
@@ -22,7 +19,7 @@ test!(add_filter_anchor {
 
     assert_matches!(pf.add_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
 
-    let anchors = pfcli::get_anchors(None).unwrap();
+    let anchors = pfcli::get_anchors(None);
     assert!(anchors.contains(&anchor_name));
 
     assert_matches!(
@@ -39,7 +36,7 @@ test!(remove_filter_anchor {
     assert_matches!(pf.add_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
     assert_matches!(pf.remove_anchor(&anchor_name, pfctl::AnchorKind::Filter), Ok(()));
 
-    let anchors = pfcli::get_anchors(None).unwrap();
+    let anchors = pfcli::get_anchors(None);
     assert!(!anchors.contains(&anchor_name));
 
     assert_matches!(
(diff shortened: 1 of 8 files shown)


dlon (approved):
:lgtm:

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/104#-:-O0NlmJlCwnLVvA93QTU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`tests/redirect_rules.rs` line 52 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/104#-O0Nks-X7iIGJwng9uDk:-O0Nks-X7iIGJwng9uDl:bqxb22k) ([raw file](https://github.com/mullvad/pfctl-rs/blob/08f1768577954d504b742e4db4b387093cbb7474/tests/redirect_rules.rs#L52)):*
> ```XML
>         assert_matches!(pf.flush_rules(ANCHOR_NAME, pfctl::RulesetKind::Redirect), Ok(()));
>         assert_eq!(
>             pfcli::get_nat_rules(ANCHOR_NAME),
> ```

Nit: I think you could replace this with `assert!(pfctl::get_nat_rules(ANCHOR_NAME).is_empty())` or  `assert_eq!(pfctl::get_nat_rules(ANCHOR_NAME).len(), 0)`.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/104)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`tests/redirect_rules.rs` line 52 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/104#-O0Nks-X7iIGJwng9uDk:-O0O-hFdCxPthwjSKU8Y:bubbhmc) ([raw file](https://github.com/mullvad/pfctl-rs/blob/08f1768577954d504b742e4db4b387093cbb7474/tests/redirect_rules.rs#L52)):*

Nit: I think you could replace this with `assert!(pfctl::get_nat_rules(ANCHOR_NAME).is_empty())` or  `assert_eq!(pfctl::get_nat_rules(ANCHOR_NAME).len(), 0)`.
</blockquote></details>

I can, but it would yield worse error messages. Then it would either just say that my bool was false, or it would say my number was not zero. With this construct the panic message will contain the rules returned (that was not an empty list).

PR #105, opened 2024-06-27
Simplify some error handling

faern:
This is in preparation of ditching `error-chain`. We had pretty tangled errors. Lots of our own errors are just stringly typed errors (`ensure!` kind of stuff).

This PR just massage away some errors.

* Types that implemented `TryCopyTo` but actually had no code path leading to an error was changed into implementations of the infallible `CopyTo`
* The internal `compare_cstr_safe` was changed to panic instead of returning an error. This because it only deals with parsing C strings returned by the system. Since the system should return C strings, it's something really bad going on with the system if they don't null-terminate their strings. I regard this as a case for panic rather than user error handling.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/105)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         # Keep MSRV in sync with rust-version in Cargo.toml as much as possible.
-        rust: [stable, beta, nightly, 1.62.0]
+        rust: [stable, beta, nightly, 1.69.0]
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 * Upgrade `ipnetwork` dependency from 0.16 to 0.20. This is a breaking change since
   `ipnetwork` is part of the public API.
 * Upgrade crate to Rust 2021 edition.
+* MSRV bumped to 1.69 due to use of `CStr::from_bytes_until_nul`.
 
 ### Removed
 * Remove `PoolAddrList::to_palist` from the public API. It should never have been exposed.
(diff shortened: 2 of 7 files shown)


dlon (approved):
:lgtm:

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/105#-:-O0OgLAKCtd9oPOz9TRi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`src/lib.rs` line 162 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/105#-O0OdOTx3XWwtW8qWCP7:-O0OdOU-CdLs-2GWR4jG:bjl129w) ([raw file](https://github.com/mullvad/pfctl-rs/blob/1e585d3a10f1b801c62b9f3665c2c644c1ae5457/src/lib.rs#L162)):*
> ```XML
>     // SAFETY: We point to the same memory region as `c_chars`,
>     // which is a valid slice, so it's guaranteed to be safe
>     let c_chars_u8 = unsafe { slice::from_raw_parts(c_chars_ptr, c_chars.len()) };
> ```

I'm curious how you feel about simply casting this from `*const [i8]` to `*const [u8]`:

```rust
let c_chars_u8 = unsafe { &*(c_chars as *const [i8] as *const [u8]) };
```

I'm sure their bits will be identical.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/105)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`src/lib.rs` line 162 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/105#-O0OdOTx3XWwtW8qWCP7:-O0Omj3-3-3BfWgo5Mm2:bwyd989) ([raw file](https://github.com/mullvad/pfctl-rs/blob/1e585d3a10f1b801c62b9f3665c2c644c1ae5457/src/lib.rs#L162)):*

I'm curious how you feel about simply casting this from `*const [i8]` to `*const [u8]`:

```rust
let c_chars_ut = unsafe { &*(c_chars as *const [i8] as *const [u8]) };
```

I'm sure their bits will be identical.
</blockquote></details>

As discussed on slack: That is not sound. because the memory layout of `&[T]` can't be assumed to be the same as `&[U]`. https://users.rust-lang.org/t/how-to-convert-i8-to-u8/16308/11

PR #106, opened 2024-06-28
Add connection state methods

dlon:
Add methods for listing and removing individual PF connection states. The existing methods (`clear_states`, and `clear_interface_states`) only allow states associated with a given anchor or interface to be removed, so these give more control.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 * **Security**: in case of vulnerabilities.
 
 ## [unreleased]
+### Added
+- Add function for listing all states created by PF anchor rules.
+- Add function for removing individual states created by PF anchor rules.
+
 ### Changed
 * Upgrade `ipnetwork` dependency from 0.16 to 0.20. This is a breaking change since
   `ipnetwork` is part of the public API.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,9 @@ pub use crate::anchor::*;
 mod ruleset;
 pub use crate::ruleset::*;
 
+mod state;
+pub use crate::state::*;
+
 mod transaction;
 pub use crate::transaction::*;
 
@@ -110,6 +113,12 @@ pub enum ErrorKind {
     InvalidPortRange,
     /// The supplied rule label is not compatible with PF.
     InvalidLabel,
+    /// The address family is invalid
+    InvalidAddressFamily,
+    /// The direction is invalid
+    InvalidDirection,
+    /// The transport protocol is invalid
+    InvalidTransportProtocol,
     /// The target state was already active
     StateAlreadyActive,
     /// This PF anchor does not exist
@@ -130,6 +139,9 @@ enum ErrorInternal {
     InvalidAnchorName(&'static str),
     InvalidPortRange,
     InvalidLabel(&'static str),
+    InvalidAddressFamily(u8),
+    InvalidDirection(u8),
+    InvalidTransportProtocol(u8),
     StateAlreadyActive,
     AnchorDoesNotExist,
     Ioctl(std::io::Error),
@@ -147,6 +159,9 @@ impl Error {
             InvalidAnchorName(..) => ErrorKind::InvalidAnchorName,
             InvalidPortRange => ErrorKind::InvalidPortRange,
             InvalidLabel(..) => ErrorKind::InvalidLabel,
+            InvalidAddressFamily(_) => ErrorKind::InvalidAddressFamily,
+            InvalidDirection(_) => ErrorKind::InvalidDirection,
+            InvalidTransportProtocol(_) => ErrorKind::InvalidTransportProtocol,
             StateAlreadyActive => ErrorKind::StateAlreadyActive,
             AnchorDoesNotExist => ErrorKind::AnchorDoesNotExist,
             Ioctl(_) => ErrorKind::Ioctl,
@@ -173,6 +188,11 @@ impl fmt::Display for Error {
             InvalidAnchorName(reason) => write!(f, "Invalid anchor name ({reason})"),
             InvalidPortRange => write!(f, "Lower port is greater than upper port"),
             InvalidLabel(reason) => write!(f, "Invalid rule label ({reason}"),
+            InvalidAddressFamily(family) => write!(f, "Invalid address family ({family})"),
+            InvalidDirection(direction) => write!(f, "Invalid direction ({direction})"),
+            InvalidTransportProtocol(protocol) => {
+                write!(f, "Invalid transport protocol ({protocol})")
+            }
             StateAlreadyActive => write!(f, "Target state is already active"),
             AnchorDoesNotExist => write!(f, "Anchor does not exist"),
             Ioctl(_) => write!(f, "Error during ioctl syscall"),
@@ -380,7 +400,7 @@ impl PfCtl {
     /// Returns total number of removed states upon success, otherwise
     /// ErrorKind::AnchorDoesNotExist if anchor does not exist.
     pub fn clear_states(&mut self, anchor_name: &str, kind: AnchorKind) -> Result<u32> {
-        let pfsync_states = self.get_states()?;
+        let pfsync_states = self.get_states_inner()?;
         if !pfsync_states.is_empty() {
             self.with_anchor_rule(anchor_name, kind, |anchor_rule| {
                 pfsync_states
@@ -414,7 +434,30 @@ impl PfCtl {
     }
 
     /// Get all states created by stateful rules
-    fn get_states(&mut self) -> Result<Vec<ffi::pfvar::pfsync_state>> {
+    pub fn get_states(&mut self) -> Result<Vec<State>> {
+        let wrapped_states = self
+            .get_states_inner()?
+            .into_iter()
+            .map(|state| {
+                // SAFETY: `state` is zero-initialized by `setup_pfioc_states`.
+                unsafe { State::new(state) }
+            })
+            .collect();
+        Ok(wrapped_states)
+    }
+
+    /// Remove the specified state.
+    ///
+    /// All current states can be obtained via [get_states].
+    pub fn kill_state(&mut self, state: &State) -> Result<()> {
+        let mut pfioc_state_kill = unsafe { mem::zeroed::<ffi::pfvar::pfioc_state_kill>() };
+        setup_pfioc_state_kill(state.as_raw(), &mut pfioc_state_kill);
+        ioctl_guard!(ffi::pf_kill_states(self.fd(), &mut pfioc_state_kill))?;
+        Ok(())
+    }
+
+    /// Get all states created by stateful rules
+    fn get_states_inner(&mut self) -> Result<Vec<ffi::pfvar::pfsync_state>> {
         let num_states = self.get_num_states()?;
         if num_states > 0 {
             let (mut pfioc_states, pfsync_states) = setup_pfioc_states(num_states);
--- a/src/rule/addr_family.rs
+++ b/src/rule/addr_family.rs
@@ -6,24 +6,21 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use crate::ffi;
+use crate::{ffi, Error, ErrorInternal, Result};
 use std::fmt;
 
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
 pub enum AddrFamily {
     #[default]
-    Any,
-    Ipv4,
-    Ipv6,
+    Any = ffi::pfvar::PF_UNSPEC as u8,
+    Ipv4 = ffi::pfvar::PF_INET as u8,
+    Ipv6 = ffi::pfvar::PF_INET6 as u8,
 }
 
 impl From<AddrFamily> for u8 {
     fn from(af: AddrFamily) -> Self {
-        match af {
-            AddrFamily::Any => ffi::pfvar::PF_UNSPEC as u8,
-            AddrFamily::Ipv4 => ffi::pfvar::PF_INET as u8,
-            AddrFamily::Ipv6 => ffi::pfvar::PF_INET6 as u8,
-        }
+        af as u8
     }
 }
 
@@ -35,5 +32,18 @@ impl fmt::Display for AddrFamily {
             AddrFamily::Ipv6 => "IPv6",
         }
         .fmt(f)
+    }
+}
+
+impl TryFrom<u8> for AddrFamily {
+    type Error = crate::Error;
+
+    fn try_from(family: u8) -> Result<Self> {
+        match family {
+            v if v == AddrFamily::Any as u8 => Ok(AddrFamily::Any),
+            v if v == AddrFamily::Ipv4 as u8 => Ok(AddrFamily::Ipv4),
+            v if v == AddrFamily::Ipv6 as u8 => Ok(AddrFamily::Ipv6),
+            _ => Err(Error::from(ErrorInternal::InvalidAddressFamily(family))),
+        }
     }
 }
--- a/src/rule/direction.rs
+++ b/src/rule/direction.rs
@@ -6,23 +6,33 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use crate::ffi;
+use crate::{ffi, Error, ErrorInternal, Result};
 
 /// Enum describing matching of rule towards packet flow direction.
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
 pub enum Direction {
     #[default]
-    Any,
-    In,
-    Out,
+    Any = ffi::pfvar::PF_INOUT as u8,
+    In = ffi::pfvar::PF_IN as u8,
+    Out = ffi::pfvar::PF_OUT as u8,
 }
 
 impl From<Direction> for u8 {
     fn from(direction: Direction) -> Self {
+        direction as u8
+    }
+}
+
+impl TryFrom<u8> for Direction {
+    type Error = crate::Error;
+
+    fn try_from(direction: u8) -> Result<Self> {
         match direction {
-            Direction::Any => ffi::pfvar::PF_INOUT as u8,
-            Direction::In => ffi::pfvar::PF_IN as u8,
-            Direction::Out => ffi::pfvar::PF_OUT as u8,
+            v if v == Direction::Any as u8 => Ok(Direction::Any),
+            v if v == Direction::In as u8 => Ok(Direction::In),
+            v if v == Direction::Out as u8 => Ok(Direction::Out),
+            other => Err(Error::from(ErrorInternal::InvalidDirection(other))),
         }
     }
 }
--- a/src/rule/proto.rs
+++ b/src/rule/proto.rs
@@ -6,14 +6,17 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use crate::{Error, ErrorInternal, Result};
+
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
 pub enum Proto {
     #[default]
-    Any,
-    Tcp,
-    Udp,
-    Icmp,
-    IcmpV6,
+    Any = libc::IPPROTO_IP as u8,
+    Tcp = libc::IPPROTO_TCP as u8,
+    Udp = libc::IPPROTO_UDP as u8,
+    Icmp = libc::IPPROTO_ICMP as u8,
+    IcmpV6 = libc::IPPROTO_ICMPV6 as u8,
 }
 
 impl From<Proto> for u8 {
@@ -24,6 +27,21 @@ impl From<Proto> for u8 {
             Proto::Udp => libc::IPPROTO_UDP as u8,
             Proto::Icmp => libc::IPPROTO_ICMP as u8,
             Proto::IcmpV6 => libc::IPPROTO_ICMPV6 as u8,
+        }
+    }
+}
+
+impl TryFrom<u8> for Proto {
+    type Error = crate::Error;
+
+    fn try_from(proto: u8) -> Result<Self> {
+        match proto {
+            v if v == Proto::Any as u8 => Ok(Proto::Any),
+            v if v == Proto::Tcp as u8 => Ok(Proto::Tcp),
+            v if v == Proto::Udp as u8 => Ok(Proto::Udp),
+            v if v == Proto::Icmp as u8 => Ok(Proto::Icmp),
+            v if v == Proto::IcmpV6 as u8 => Ok(Proto::IcmpV6),
+            _ => Err(Error::from(ErrorInternal::InvalidTransportProtocol(proto))),
         }
     }
 }
--- a/src/state.rs
+++ b/src/state.rs
@@ -87,3 +87,40 @@ unsafe fn parse_address(family: u8, host: pfsync_state_host) -> Result<SocketAdd
 
     Ok(SocketAddr::new(ip, port))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::pfsync_state_host;
+    use crate::{state::parse_address, AddrFamily};
+    use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
+
+    #[test]
+    fn test_parse_ipv4_address() {
+        const EXPECTED_IP: Ipv4Addr = Ipv4Addr::new(1, 2, 3, 4);
+        const EXPECTED_PORT: u16 = 12345;
+
+        let mut host: pfsync_state_host = unsafe { std::mem::zeroed() };
+        host.addr.pfa._v4addr.s_addr = u32::from_be_bytes(EXPECTED_IP.octets()).to_be();
+        host.xport.port = EXPECTED_PORT.to_be();
+
+        let family = u8::from(AddrFamily::Ipv4);
+
+        let address = unsafe { parse_address(family, host) }.unwrap();
+        assert_eq!(address, SocketAddr::new(EXPECTED_IP.into(), EXPECTED_PORT));
+    }
+
+    #[test]
+    fn test_parse_ipv6_address() {
+        const EXPECTED_IP: Ipv6Addr = Ipv6Addr::new(1, 2, 3, 4, 5, 6, 7, 0x7f);
+        const EXPECTED_PORT: u16 = 12345;
+
+        let mut host: pfsync_state_host = unsafe { std::mem::zeroed() };
+        host.addr.pfa._v6addr.__u6_addr.__u6_addr8 = EXPECTED_IP.octets();
+        host.xport.port = EXPECTED_PORT.to_be();
+
+        let family = u8::from(AddrFamily::Ipv6);
+
+        let address = unsafe { parse_address(family, host) }.unwrap();
+        assert_eq!(address, SocketAddr::new(EXPECTED_IP.into(), EXPECTED_PORT));
+    }
+}
--- a/src/state.rs
+++ b/src/state.rs
@@ -1,0 +1,89 @@
+use std::fmt;
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
+
+use crate::ffi::pfvar::pfsync_state_host;
+use crate::{ffi::pfvar::pfsync_state, Direction, Proto};
+use crate::{Error, ErrorInternal, Result};
+
+/// PF connection state created by a stateful rule
+#[derive(Clone)]
+pub struct State {
+    sync_state: pfsync_state,
+}
+
+// Manually derive `Debug` since `pfsync_state` contains unions.
+impl fmt::Debug for State {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("State")
+            .field("direction", &self.direction())
+            .field("proto", &self.proto())
+            .field("local_address", &self.local_address())
+            .field("remote_address", &self.remote_address())
+            .finish()
+    }
+}
+
+impl State {
+    /// Wrap a `pfsync_state` so that it can be accessed safely.
+    ///
+    /// # Safety
+    ///
+    /// All bytes in `sync_state` must be initialized.
+    pub(crate) unsafe fn new(sync_state: pfsync_state) -> State {
+        State { sync_state }
+    }
+
+    /// Return the direction for this state
+    pub fn direction(&self) -> Result<Direction> {
+        Direction::try_from(self.sync_state.direction)
+    }
+
+    /// Return the transport protocol for this state
+    pub fn proto(&self) -> Result<Proto> {
+        Proto::try_from(self.sync_state.proto)
+    }
+
+    /// Return the local socket address for this state
+    pub fn local_address(&self) -> Result<SocketAddr> {
+        // SAFETY: The address and port are initialized according to the contract of `Self::new`.
+        unsafe { parse_address(self.sync_state.af_lan, self.sync_state.lan) }
+    }
+
+    /// Return the remote socket address for this state
+    pub fn remote_address(&self) -> Result<SocketAddr> {
+        // SAFETY: The address and port are initialized according to the contract of `Self::new`.
+        unsafe { parse_address(self.sync_state.af_lan, self.sync_state.ext_lan) }
+    }
+
+    /// Return a reference to the inner `pfsync_state` state
+    pub(crate) fn as_raw(&self) -> &pfsync_state {
+        &self.sync_state
+    }
+}
+
+/// Parse an IP address and port from a `pfsync_sync_host`, normally provided by `pfsync_state`.
+///
+/// # Safety
+///
+/// `host` must contain a valid address and a port:
+/// * If `family == PF_INET`, then `host.addr.pfa._v4addr` must be initialized.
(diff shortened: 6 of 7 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`src/state.rs` line 15 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O0m9_xR1_eSTtf2dR_X:-O0m9_xR1_eSTtf2dR_Y:b-ebmfma) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0cae58767a64a3859cbf14f7a41842c46135ac82/src/state.rs#L15)):*
> ```XML
>     Ip(IpState),
>     /// Any connection state that could not be parsed
>     Raw(pfsync_state),
> ```

This publicly exposes a type from the `ffi` module. We can't do this. Everything related to the FFI and automatically generated bindings are *internal details* only.
___
*[`src/state.rs` line 32 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O0mAdetE_0FT0gD-355:-O0mAdeuCrMT9YG9_-GK:b6vxo6p) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0cae58767a64a3859cbf14f7a41842c46135ac82/src/state.rs#L32)):*
> ```XML
>         IpState::try_from(state)
>             .map(State::Ip)
>             .unwrap_or_else(|_| State::Raw(state))
> ```

Is there no field in `pfsync_state` that indicate what kind of state it is? Having it be raw just because it fails to parse as an IP state feels slightly off, but I don't know the PF internals around this.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`src/lib.rs` line 344 at r1](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O0mBOSH7iL50amIDOGd:-O0mBOSH7iL50amIDOGe:b-h1sn2r) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0cae58767a64a3859cbf14f7a41842c46135ac82/src/lib.rs#L344)):*
I think the documentation would be clearer if it mentions that it *removes* states for which the filter returns `false`. Because that's the "action" that is taken in this method. Nothing actually happens to the states where the filter returns true, they are just ignored.

> Keep all states for which `filter` returns true. The rest are removed from PF.

?

Would also be nice if it mentions that if the function returns an error, some states *might* have been removed already.

> If an error is returned, all states except the last one for which the filter returned `false` has already been removed.

faern (commented):
> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @dlon)

___
*[`src/lib.rs` line 92 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UIUUN2Ga-qzfx1z4a:-O2UIUUO3VoagGXg_WBU:bisvqsh) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L92)):*
> ```XML
> 
> mod state;
> pub use crate::state::*;
> ```

The `state` module only has a single type. So this is fine. But otherwise I suggest we stop re-exporting everything from all modules like this. The main namespace of this crate is growing pretty big. Having a bit of a tree structure would probably clean it up a bit. But again, not important here since it's only one item.
___
*[`src/lib.rs` line 188 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UIG9A9Yv2I00IR6J_:-O2UIG9A9Yv2I00IR6Ja:bv7l7bj) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L188)):*
> ```XML
>             InvalidAddressFamily(reason) => write!(f, "Invalid address family ({reason})"),
>             InvalidDirection(reason) => write!(f, "Invalid direction ({reason})"),
>             InvalidTransportProtocol(reason) => write!(f, "Invalid transport protocol ({reason})"),
> ```

I think we should use a better name than `reason` here. These are not reasons. These are `address_family`, `direction` and `protocol` respectively.
___
*[`src/lib.rs` line 436 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UJ8N60x9X91nHHbkF:-O2UJ8N60x9X91nHHbkG:bwrj62k) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L436)):*
> ```XML
>         setup_pfioc_state_kill(state.as_raw(), &mut pfioc_state_kill);
>         ioctl_guard!(ffi::pf_kill_states(self.fd(), &mut pfioc_state_kill))?;
>         // psk_af holds the number of killed states, but it should be zero
> ```

Why zero? Would it not match zero or one states? Zero if the state was already gone at the time of issuing this call, or one if it was removed?
___
*[`src/lib.rs` line 445 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UJf3UEH8s9HhtR4XI:-O2UJf3UEH8s9HhtR4XJ:b-yrbj9w) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L445)):*
> ```XML
>         if num_states > 0 {
>             let (mut pfioc_states, pfsync_states) = setup_pfioc_states(num_states);
>             ioctl_guard!(ffi::pf_get_states(self.fd(), &mut pfioc_states))?;
> ```

Does this have a race? What happens if the number of states changes between the call to get number of states and the syscall to get the states?

Not fully related to this PR, but I noticed it now. Can we improve it to have retry logic or in some other way not fail in this way? Once we start using this feature to list and kill states we will be calling this method much more often, and be more likely to trigger the race, if there is one.
___
*[`src/state.rs` line 8 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UOqHP0L3UlywQ1UMQ:-O2UOqHP0L3UlywQ1UMR:b-py2dut) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L8)):*
> ```XML
> 
> /// PF connection state created by a stateful rule
> #[derive(Clone)]
> ```

`Debug` would be nice. I'm not sure how it would be automatically implemented, but good practice to have `Debug` on all public types.
___
*[`src/state.rs` line 25 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UP2qz8YzicXbiHu-M:-O2UP2qz8YzicXbiHu-N:bx1rw3c) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L25)):*
> ```XML
>     /// Return the transport protocol for this state
>     pub fn proto(&self) -> Result<Proto> {
>         Proto::try_from(self.sync_state.direction)
> ```

Typo? You pass in `sync_state.direction` and not the proto. Something we could catch with a unit test! Can you consider adding unit tests for this type? That would also help to make sure the endianness ninjaing in `parse_address` is correct and stays correct during refactoring.
___
*[`src/state.rs` line 54 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UPrVIBXZ_Jf01Tyvk:-O2UPrVIBXZ_Jf01Tyvl:b-snugr5) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L54)):*
> ```XML
>         _ => return Err(Error::from(ErrorInternal::InvalidAddressFamily(family))),
>     };
>     let port = unsafe { host.xport.port }.to_be();
> ```

Should this not be `from_be`? :thinking:  Is `host.xport.port` not stored in big endian? `SocketAddr::new` below wants the port in native-endianness.
___
*[`src/rule/direction.rs` line 36 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UKVvJC8pjNCDyk0cF:-O2UKVvJC8pjNCDyk0cG:bziikaq) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/rule/direction.rs#L36)):*
> ```XML
>         const INOUT: u8 = ffi::pfvar::PF_INOUT as u8;
>         const IN: u8 = ffi::pfvar::PF_IN as u8;
>         const OUT: u8 = ffi::pfvar::PF_OUT as u8;
> ```

We should probably not duplicate these. They are already defined and casted above.

Can we set `#[repr(u8)]` on the enum and assign these constants directly on the Rust variants? Then we can simply use our own enum when converting/matching.

Same goes for the other enums you implement `TryFrom<u8>` in this PR. Now when we go both ways, we should probably find a way to not duplicate a lot of FFI constants and their casting. Since your PR goes from needing them in a single place to two places, this work sort of lands on this PR.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 4 of 5 files reviewed, 8 unresolved discussions (waiting on @dlon)

___
*[`src/lib.rs` line 435 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2US2GEA2Z_NR4BxWSE:-O2US2GEA2Z_NR4BxWSF:b-89opfm) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L435)):*
> ```XML
>         let mut pfioc_state_kill = unsafe { mem::zeroed::<ffi::pfvar::pfioc_state_kill>() };
>         setup_pfioc_state_kill(state.as_raw(), &mut pfioc_state_kill);
>         ioctl_guard!(ffi::pf_kill_states(self.fd(), &mut pfioc_state_kill))?;
> ```

Was it possible to use some kind of ID from the state? To ensure we only remove one state, and not all matching states? :thinking:

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 2 of 11 files reviewed, 8 unresolved discussions (waiting on @faern)

___
*[`src/lib.rs` line 435 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2US2GEA2Z_NR4BxWSE:-O2UXswu104AvB_A4AKo:b-i3hnma) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L435)):*

Was it possible to use some kind of ID from the state? To ensure we only remove one state, and not all matching states? :thinking:
</blockquote></details>

I looked into it and DIOCKILLSTATES does not check any kind of state ID: https://github.com/apple/darwin-xnu/blob/main/bsd/net/pf_ioctl.c#L3403

But I believe a `(interface, protocol, local IP, local port, remote IP, remote port)` tuple will be unique anyway?
___
*[`src/lib.rs` line 436 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UJ8N60x9X91nHHbkF:-O2VNtam8hqfG7bXKojv:b423gd2) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L436)):*

Why zero? Would it not match zero or one states? Zero if the state was already gone at the time of issuing this call, or one if it was removed?
</blockquote></details>

I removed the comment, since we don't even refer to the field.
___
*[`src/state.rs` line 8 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UOqHP0L3UlywQ1UMQ:-O2VO0gV8iftWoMymBWC:b-vw7nk5) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L8)):*

`Debug` would be nice. I'm not sure how it would be automatically implemented, but good practice to have `Debug` on all public types.
</blockquote></details>

Unfortunately, `pfsync_state` contains unions, so the `Debug` macro didn't work. I reran bindgen with `--disable-untagged-union`. But the output is not particularly useful, IMO. Maybe just manually implementing `Debug` would be fine?
___
*[`src/state.rs` line 54 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UPrVIBXZ_Jf01Tyvk:-O2UjzRb5VCHfu0S1pOB:b-3hz2h7) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L54)):*

Should this not be `from_be`? :thinking:  Is `host.xport.port` not stored in big endian? `SocketAddr::new` below wants the port in native-endianness.
</blockquote></details>

You're right! Fixed.

faern (commented):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 3 of 11 files reviewed, 6 unresolved discussions (waiting on @dlon)

___
*[`src/lib.rs` line 435 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2US2GEA2Z_NR4BxWSE:-O2WDSBYCigUx6GKjPnR:bu7f564) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L435)):*

I looked into it and DIOCKILLSTATES does not check any kind of state ID: https://github.com/apple/darwin-xnu/blob/main/bsd/net/pf_ioctl.c#L3403

But I believe a `(interface, protocol, local IP, local port, remote IP, remote port)` tuple will be unique anyway?
</blockquote></details>

Ok. Maybe it will. I have not dived into the details. Given your question mark you don't sound sure either :wink: But yeah, at least I guess we can assume only states matching all the fields. If there happens to be >1 state matching then at least the user already told us to remove those state(s). So I guess it's fine. Let's move on!
___
*[`src/lib.rs` line 431 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WE_-P4BRnHSwKojO1:-O2WE_-P4BRnHSwKojO2:b3tyqle) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/lib.rs#L431)):*
Please add information on how to obtain a `&State`. Otherwise it's hard to know how to use this method. Something like:

> /// Remove the specified state.
> ///
> /// All current states can be obtained via [`get_states`].
___
*[`src/state.rs` line 8 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UOqHP0L3UlywQ1UMQ:-O2WFn8X1NXHwTUYfyf9:b-d87vkh) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L8)):*

Unfortunately, `pfsync_state` contains unions, so the `Debug` macro didn't work. I reran bindgen with `--disable-untagged-union`. But the output is not particularly useful, IMO. Maybe just manually implementing `Debug` would be fine?
</blockquote></details>

I think we should revert the changes to the FFI. It became a very large change, and primarily just more complicated types. Either leave `Debug` out, or implement it manually IMO.
___
*[`src/state.rs` line 56 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHNnS7IBR-_OgX0JD:-O2WHNnS7IBR-_OgX0JE:bc4fbb8) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/state.rs#L56)):*
> ```XML
>         _ => return Err(Error::from(ErrorInternal::InvalidAddressFamily(family))),
>     };
>     let port = u16::from_be(unsafe { *host.xport.port.as_ref() });
> ```

This function has a lot of `unsafe`. But no safety docs. Is it possible to cause UB by calling this function with carefully crafted bad arguments? If so, the function needs to be `unsafe`.

faern (commented):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 4 of 11 files reviewed, 7 unresolved discussions (waiting on @dlon)

___
*[`generate_bindings.sh` line 25 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHutpAowDADC7BJin:-O2WHutpAowDADC7BJio:b-xklmme) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/generate_bindings.sh#L25)):*
> ```Shell
>     --allowlist-var PFRULE_.* \
>     --default-enum-style rust \
>     --disable-untagged-union \
> ```

I don't think we should change this unless we have a good reason. Currently it mostly looks like it messes a lot with the library code, for very little gain, if any.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @faern)

___
*[`src/lib.rs` line 92 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UIUUN2Ga-qzfx1z4a:-O2Z5jJj3lElAeCJbxFR:bf6lylp) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L92)):*

The `state` module only has a single type. So this is fine. But otherwise I suggest we stop re-exporting everything from all modules like this. The main namespace of this crate is growing pretty big. Having a bit of a tree structure would probably clean it up a bit. But again, not important here since it's only one item.
</blockquote></details>

Agreed that we should not do this in general, but I think it's more important to be consistent. This would be a breaking change, though.
___
*[`src/lib.rs` line 188 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UIG9A9Yv2I00IR6J_:-O2YiDEC0m3Jd1HVRcqv:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L188)):*

I think we should use a better name than `reason` here. These are not reasons. These are `address_family`, `direction` and `protocol` respectively.
</blockquote></details>

Done.
___
*[`src/lib.rs` line 445 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UJf3UEH8s9HhtR4XI:-O2Z4iXAD4J-fq3_6Ujp:b-67pm7l) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L445)):*

Does this have a race? What happens if the number of states changes between the call to get number of states and the syscall to get the states?

Not fully related to this PR, but I noticed it now. Can we improve it to have retry logic or in some other way not fail in this way? Once we start using this feature to list and kill states we will be calling this method much more often, and be more likely to trigger the race, if there is one.
</blockquote></details>

Yeah, this is definitely wrong! Retrying while the buffer is too small is also what `pfctl` does: https://github.com/openbsd/src/blob/master/sbin/pfctl/pfctl.c#L1047
___
*[`src/state.rs` line 8 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UOqHP0L3UlywQ1UMQ:-O2YjESFDjXr2Ba2aV_D:b-4ni3ju) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L8)):*

I think we should revert the changes to the FFI. It became a very large change, and primarily just more complicated types. Either leave `Debug` out, or implement it manually IMO.
</blockquote></details>

Added a manual implementation.
___
*[`src/state.rs` line 25 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UP2qz8YzicXbiHu-M:-O2VOcJpDUOYeE62iZS7:bppesd6) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L25)):*

Typo? You pass in `sync_state.direction` and not the proto. Something we could catch with a unit test! Can you consider adding unit tests for this type? That would also help to make sure the endianness ninjaing in `parse_address` is correct and stays correct during refactoring.
</blockquote></details>

This is now covered both by unit tests and the integration tests I've added.
___
*[`src/state.rs` line 56 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHNnS7IBR-_OgX0JD:-O2YvQXgAznlbbPqT0Z1:b1h3tc4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/state.rs#L56)):*

This function has a lot of `unsafe`. But no safety docs. Is it possible to cause UB by calling this function with carefully crafted bad arguments? If so, the function needs to be `unsafe`.
</blockquote></details>

It should always be safe, because the struct is zero-initialized. But it's not always meaningful. Commented to clarify.
___
*[`src/rule/direction.rs` line 36 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UKVvJC8pjNCDyk0cF:-O2Z48FxEw9a2zJQEou3:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/rule/direction.rs#L36)):*

We should probably not duplicate these. They are already defined and casted above.

Can we set `#[repr(u8)]` on the enum and assign these constants directly on the Rust variants? Then we can simply use our own enum when converting/matching.

Same goes for the other enums you implement `TryFrom<u8>` in this PR. Now when we go both ways, we should probably find a way to not duplicate a lot of FFI constants and their casting. Since your PR goes from needing them in a single place to two places, this work sort of lands on this PR.
</blockquote></details>

Done.
___
*[`generate_bindings.sh` line 25 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHutpAowDADC7BJin:-O2YhUpN75fDF5Fdrlex:bp3aow6) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/generate_bindings.sh#L25)):*

I don't think we should change this unless we have a good reason. Currently it mostly looks like it messes a lot with the library code, for very little gain, if any.
</blockquote></details>

Reverted.

faern (commented):
> Reviewed 10 of 11 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`src/lib.rs` line 445 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UJf3UEH8s9HhtR4XI:-O2ZJGh36Y9Z93dB4nVO:b-mly9gh) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/lib.rs#L445)):*

Yeah, this is definitely wrong! Retrying while the buffer is too small is also what `pfctl` does: https://github.com/openbsd/src/blob/master/sbin/pfctl/pfctl.c#L1047
</blockquote></details>

Ok. Let's leave that for a separate PR! But this needs fixing. I created an issue: https://github.com/mullvad/pfctl-rs/issues/109
___
*[`src/state.rs` line 25 at r2](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2UP2qz8YzicXbiHu-M:-O2ZMyEL40XmFR03L5Fz:b-ethmss) ([raw file](https://github.com/mullvad/pfctl-rs/blob/0809447e6c2879df2940a4a14dff41394566db76/src/state.rs#L25)):*

This is now covered both by unit tests and the integration tests I've added.
</blockquote></details>

Really nice!
___
*[`src/state.rs` line 56 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHNnS7IBR-_OgX0JD:-O2ZOUR708SqOsayNcN9:b27s2jl) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/state.rs#L56)):*
> because the struct is zero-initialized

This is not something this function guarantees. A function's safety contract has to be upheld by that function itself, not how it's called by others.

Maybe this is safe... What guarantees does Rust give about the unused memory of a union?

If I have
```rust
union Foo {
    small: u8,
    large: u128,
}
```
And I initialize a `Foo { small: 0 }`. Is it then safe to read `foo.large`? If so, this function is safe. Otherwise not. Because here we read `host.addr.pfa._v6addr`. This is not the smallest field in the enum, so what happens when someone calls `parse_address` with a `pfa` initialized to a smaller union variant than v6?
___
*[`src/rule/direction.rs` line 18 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2ZLSlpCPpFXJ7VpfP8:-O2ZLSlpCPpFXJ7VpfP9:bnpuzns) ([raw file](https://github.com/mullvad/pfctl-rs/blob/6d40bf7cbcbd484d201048a7a5be5bfaeed92501/src/rule/direction.rs#L18)):*
> ```XML
>     Any = ffi::pfvar::PF_INOUT as u32 as u8,
>     In = ffi::pfvar::PF_IN as u32 as u8,
>     Out = ffi::pfvar::PF_OUT as u32 as u8,
> ```

Why must this be cast twice? The code that is removed in this PR did not have to do the double cast :thinking: 
___
*[`tests/states.rs` line 161 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2ZMYcPA-i7uczoXPvX:-O2ZMYcPA-i7uczoXPvY:b12wput) ([raw file](https://github.com/mullvad/pfctl-rs/blob/6d40bf7cbcbd484d201048a7a5be5bfaeed92501/tests/states.rs#L161)):*
> ```XML
>         .expect("Could not obtain states")
>         .into_iter()
>         .filter_map(|state| ExpectedState::try_from(state).ok()).collect::<Vec<_>>();
> ```

This construct is nice! May I suggest you use it above in the previous `pf.get_states()` also? That makes the code inside the for loop simpler (currently pretty hard to read it all since it does a lot in one go)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`src/state.rs` line 56 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WHNnS7IBR-_OgX0JD:-O2ZSaVmDP_v-DfZxq65:b-52jlxt) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/state.rs#L56)):*

> because the struct is zero-initialized

This is not something this function guarantees. A function's safety contract has to be upheld by that function itself, not how it's called by others.

Maybe this is safe... What guarantees does Rust give about the unused memory of a union?

If I have
```rust
union Foo {
    small: u8,
    large: u128,
}
```
And I initialize a `Foo { small: 0 }`. Is it then safe to read `foo.large`? If so, this function is safe. Otherwise not. Because here we read `host.addr.pfa._v6addr`. This is not the smallest field in the enum, so what happens when someone calls `parse_address` with a `pfa` initialized to a smaller union variant than v6?
</blockquote></details>

Valgrind does *not* like the following program:

```rust
fn main() {
    let foo = Foo { small: 123 };
    println!("{}", unsafe { foo.large });
}

union Foo {
    small: u8,
    large: u128,
}
```
But it does not trigger any errors if `foo` is initialized as `Foo { large: 123 };`. So yeah, The current implementation of `parse_address` *is unsafe* as far as I can tell.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`src/state.rs` line 44 at r6](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2Zcv2BBj1y89bcOBX3:-O2Zcv2BBj1y89bcOBX4:b-4pldc7) ([raw file](https://github.com/mullvad/pfctl-rs/blob/85267af323109d678bdc6addd0c395445d4d512d/src/state.rs#L44)):*
Good documentation on `parse_address`!

But now this method makes a lot of assumptions that has nothing with `local_address` to do. When you (as discussed AFK) add unsafety to `State::new` and explain why it's unsafe, then this safety documentation can be something like:

> If the safety contract in `State::new` is upheld, `sync_state.lan` is guaranteed to be initialized according to the `sync_state.af_lan` family.
___
*[`src/state.rs` line 131 at r6](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2Zc_Mo8Mg_qBTsMFYE:-O2Zc_Mo8Mg_qBTsMFYF:bj9yg38) ([raw file](https://github.com/mullvad/pfctl-rs/blob/85267af323109d678bdc6addd0c395445d4d512d/src/state.rs#L131)):*
> ```XML
>         let family = u8::from(AddrFamily::Ipv6);
> 
>         assert_matches!(unsafe { parse_address(family, host) }, Ok(addr) if addr == SocketAddr::new(EXPECTED_IP.into(), EXPECTED_PORT));
> ```

Nit. But instead of making this line super long. I would probably have a line before it with `let address = unsafe { parse_address(family, host) }.unwrap();`. This splits the complexity into two statements.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: 8 of 11 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`src/lib.rs` line 431 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2WE_-P4BRnHSwKojO1:-O2ZaQh-1MZ0UroyRPRh:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/d4d2b7683d072b07e09b0bb85565081f004f0103/src/lib.rs#L431)):*

Please add information on how to obtain a `&State`. Otherwise it's hard to know how to use this method. Something like:

> /// Remove the specified state.
> ///
> /// All current states can be obtained via [`get_states`].
</blockquote></details>

Done.
___
*[`src/state.rs` line 44 at r6](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2Zcv2BBj1y89bcOBX3:-O2Zhc5uCvjcaNfrbJK6:b-vi7nsk) ([raw file](https://github.com/mullvad/pfctl-rs/blob/85267af323109d678bdc6addd0c395445d4d512d/src/state.rs#L44)):*

Good documentation on `parse_address`!

But now this method makes a lot of assumptions that has nothing with `local_address` to do. When you (as discussed AFK) add unsafety to `State::new` and explain why it's unsafe, then this safety documentation can be something like:

> If the safety contract in `State::new` is upheld, `sync_state.lan` is guaranteed to be initialized according to the `sync_state.af_lan` family.
</blockquote></details>

Done. Would it be acceptable to simply state that the object passed to `new` must be zero-initialized? I think that covers all cases without being overly verbose.
___
*[`src/state.rs` line 131 at r6](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2Zc_Mo8Mg_qBTsMFYE:-O2Zj9ih38eKmTDKjetm:b-2vryu4) ([raw file](https://github.com/mullvad/pfctl-rs/blob/85267af323109d678bdc6addd0c395445d4d512d/src/state.rs#L131)):*

Nit. But instead of making this line super long. I would probably have a line before it with `let address = unsafe { parse_address(family, host) }.unwrap();`. This splits the complexity into two statements.
</blockquote></details>

Done. Much better.
___
*[`src/rule/direction.rs` line 18 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2ZLSlpCPpFXJ7VpfP8:-O2ZiadaDBABWV3wDBYQ:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/6d40bf7cbcbd484d201048a7a5be5bfaeed92501/src/rule/direction.rs#L18)):*

Why must this be cast twice? The code that is removed in this PR did not have to do the double cast :thinking:
</blockquote></details>

Done.
___
*[`tests/states.rs` line 161 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2ZMYcPA-i7uczoXPvX:-O2ZidjT5xcIc0ceotE0:bpnqdp1) ([raw file](https://github.com/mullvad/pfctl-rs/blob/6d40bf7cbcbd484d201048a7a5be5bfaeed92501/tests/states.rs#L161)):*

This construct is nice! May I suggest you use it above in the previous `pf.get_states()` also? That makes the code inside the for loop simpler (currently pretty hard to read it all since it does a lot in one go)
</blockquote></details>

I didn't do that since the original state is needed for `kill_state`.

faern (approved):
> Reviewed 2 of 3 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/106)* status: all files reviewed, 1 unresolved discussion

___
*[`src/state.rs` line 44 at r6](https://reviewable.io/reviews/mullvad/pfctl-rs/106#-O2Zcv2BBj1y89bcOBX3:-O2ZluQpEYlhKV5jSY4U:bhudbcw) ([raw file](https://github.com/mullvad/pfctl-rs/blob/85267af323109d678bdc6addd0c395445d4d512d/src/state.rs#L44)):*

Done. Would it be acceptable to simply state that the object passed to `new` must be zero-initialized? I think that covers all cases without being overly verbose.
</blockquote></details>

Much simpler! Let's keep it simple :P

PR #107, opened 2024-07-12
Remove error-chain and replace with handcrafted errors

faern:
This changes the API of the error a lot, and the returned error chain in many calls into this library.

This supersedes #88. I'm not saying we absolutely have to stay away from `thiserror`. But I had some questions about the details in that PR so I decided to try my own approach. And I like this approach. The error was easy to implement manually.

I have tried these changes out in our main app, and it does not create any trouble. We don't really match against the errors anywhere, so it just compiles with no changes. The error chain will print a bit differently if we run into an error, but hopefully the changes or not for the worse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/107)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   `ipnetwork` is part of the public API.
 * Upgrade crate to Rust 2021 edition.
 * MSRV bumped to 1.69 due to use of `CStr::from_bytes_until_nul`.
+* Replace `error-chain` generated errors with manually implemented error types.
 
 ### Removed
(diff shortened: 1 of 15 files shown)


Serock3 (approved):
Nice work! I like your implementation, although I probably would have preferred to use `thiserror` myself because of its simplicity, ubiquity and the fact that it would produce a very similar end product. Since you've already written it by hand I see no reason to change it, however :lgtm: 

> Reviewed 12 of 12 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/107#-:-O2A5WxiD7pR55FHJkvi:bxv5fbq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #117, opened 2024-11-01
Add `user` and `group` to NatRule

hulthe:


hulthe:
Closing because user and group don't seem to matter for NAT rules.

PR #129, opened 2025-09-16
Add function for getting interface flags

### LINKED ISSUE #128: Add way to obtain interface flags
Hi,

Thanks for releasing `v0.7.0`. It seems that while there is a way to set or clear interface flags, the corresponding method to obtain flags currently set is missing.

I suppose implementing [DIOCIGETIFACES](https://man.openbsd.org/pf#DIOCIGETIFACES) could be the way to obtain flags.

pronebird:
Closes #128

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)


pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`src/lib.rs` line 534 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_Lrz9BEP8xXOi6QkBW:-O_Lrz9BEP8xXOi6QkBX:b-jmz90h) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/lib.rs#L534)):*
> ```XML
>     ///
>     /// https://man.freebsd.org/cgi/man.cgi?pf(4)
>     pub fn get_interface_flags(
> ```

Nit : maybe this should be renamed to `get_interfaces`?

hulthe (changes requested):
> @hulthe reviewed 7 of 9 files at r1, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

___
*[`src/lib.rs` line 564 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_HgqRt2T_shXtdzUpb:-O_pmyS39xOQXlbiqyvW:bdg6zhw) ([raw file](https://github.com/mullvad/pfctl-rs/blob/adde68e8cf401f12be8cab78fee1886a221ca578/src/lib.rs#L564)):*
<details><summary><i>Previously, pronebird (Andrej Mihajlov) wrote…</i></summary><blockquote>

- Addressed with loop and simple counter.
- Fixed tests to reset SKIP flag because it seems that some of tun interfaces are being reused when running the same tests multiple times causing SKIP flag to appear in other tests
</blockquote></details>

Why have MAX_RETRIES at all? Do you foresee a risk that `pfiio_size` would grow forever?
___
*[`src/lib.rs` line 534 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_Lrz9BEP8xXOi6QkBW:-Oa-2aH_8lA3j8gkABwf:bowo3b3) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/lib.rs#L534)):*
<details><summary><i>Previously, pronebird (Andrej Mihajlov) wrote…</i></summary><blockquote>

Nit : maybe this should be renamed to `get_interfaces`?
</blockquote></details>

Yes, I agree
___
*[`src/rule/interface.rs` line 54 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-Oa-6sCl9Ee5PCUc7075:-Oa-6sCl9Ee5PCUc7076:blked3u) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/rule/interface.rs#L54)):*
> ```XML
>         const SKIP = 0x0100;
>     }
> }
> ```

Nice, i like the use of bitflags here.

This is a breaking change though, so it should be mentioned in the changelog.
___
*[`src/rule/interface.rs` line 68 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-Oa-5aKc3dnwlVaM1fWR:-Oa-5aKc3dnwlVaM1fWS:b-49l03q) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/rule/interface.rs#L68)):*
> ```XML
>         let c_chars_ptr = kif.pfik_name.as_ptr() as *const u8;
>         // SAFETY: valid as long as `kif` is within the scope
>         let c_chars_u8 = unsafe { std::slice::from_raw_parts(c_chars_ptr, kif.pfik_name.len()) };
> ```

This unsafe is unnecessary, please consider something like this instead:

```rust
    let name = pfik_name.map(|b: c_char| b as u8).to_vec();
    let name = CString::from_vec_with_nul(name)
        .map_err(/* etc */)
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: all files reviewed, 4 unresolved discussions (waiting on @hulthe)

___
*[`src/lib.rs` line 564 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_HgqRt2T_shXtdzUpb:-Oa-NeK1ApJcS91rz9ee:b-n5agnu) ([raw file](https://github.com/mullvad/pfctl-rs/blob/adde68e8cf401f12be8cab78fee1886a221ca578/src/lib.rs#L564)):*
<details><summary><i>Previously, hulthe (Joakim Hulthe) wrote…</i></summary><blockquote>

Why have MAX_RETRIES at all? Do you foresee a risk that `pfiio_size` would grow forever?
</blockquote></details>

Because we don't know how much data we need in the first place and there is no call to query that. For that reason I use high enough `INITIAL_CAPACITY`.

In practice, if you set capacity, say, to `1` and run `ffi::pf_get_ifaces()`, you'll get the buffered filled with exactly one interface. However `iface.pfiio_size` will contain the actual number of interfaces registered with pf. In which case if you'd like to retrieve all of them you need to reserve a larger buffer and then re-run the syscall.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: all files reviewed, 4 unresolved discussions (waiting on @hulthe)

___
*[`src/rule/interface.rs` line 68 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-Oa-5aKc3dnwlVaM1fWR:-Oa-P0y-0RKL6o492lXr:b2jvkkg) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/rule/interface.rs#L68)):*
<details><summary><i>Previously, hulthe (Joakim Hulthe) wrote…</i></summary><blockquote>

This unsafe is unnecessary, please consider something like this instead:

```rust
    let name = pfik_name.map(|b: c_char| b as u8).to_vec();
    let name = CString::from_vec_with_nul(name)
        .map_err(/* etc */)
```
</blockquote></details>

That won't work. `pfik_name` is defined as a fixed array:

```rs
pub pfik_name: [::std::os::raw::c_char; 16usize],
```

- If interface name is 4 letters long, the rest of the `pfik_name` buffer will be filled with zeroes.
- `CString::from_vec_with_nul` will error if `0` byte is not the last one.

We could however use `CStr::from_bytes_until_nul` instead of `CString::from_vec_with_nul` to achieve what you want. However note that conversion from `i8` to `u8` is in theory fallible. When we use `std::slice::from_raw_parts` we don't manipulate bytes directly.

I don't think `unsafe` itself is a problem. If you insist I can map bytes, but I thought it's uglier than manipulating the pointer. Also this exact piece of code is already used in `compare_cstr_safe()` in the codebase so that's why I used it, as I thought it was more elegant.

FYI I have raised this ergonomics issue in the Rust repository: https://github.com/rust-lang/rust/issues/146836

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @hulthe)

___
*[`src/lib.rs` line 534 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_Lrz9BEP8xXOi6QkBW:-Oa-U4CW4mACvdB2CdgX:b-896fix) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/lib.rs#L534)):*
<details><summary><i>Previously, hulthe (Joakim Hulthe) wrote…</i></summary><blockquote>

Yes, I agree
</blockquote></details>

Done.

hulthe (changes requested):
> @hulthe reviewed 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`src/lib.rs` line 564 at r3](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-O_HgqRt2T_shXtdzUpb:-Oa-fPOz7OUiIgSn9VJo:bw83lqe) ([raw file](https://github.com/mullvad/pfctl-rs/blob/adde68e8cf401f12be8cab78fee1886a221ca578/src/lib.rs#L564)):*
<details><summary><i>Previously, pronebird (Andrej Mihajlov) wrote…</i></summary><blockquote>

Because we don't know how much data we need in the first place and there is no call to query that. For that reason I use high enough `INITIAL_CAPACITY`.

In practice, if you set capacity, say, to `1` and run `ffi::pf_get_ifaces()`, you'll get the buffered filled with exactly one interface. However `iface.pfiio_size` will contain the actual number of interfaces registered with pf. In which case if you'd like to retrieve all of them you need to reserve a larger buffer and then re-run the syscall.
</blockquote></details>

I'm not asking why the loop exists. I'm asking why you limit the loop to 2 iterations since the second call to `pf_get_ifaces` might, in an racy scenario, still return a `pfiio_size` that is larger than `buf.capacity()`.

But as I'm writing this I realize it's impossible to not be racy anyway, so please ignore me :)
___
*[`src/rule/interface.rs` line 68 at r5](https://reviewable.io/reviews/mullvad/pfctl-rs/129#-Oa-5aKc3dnwlVaM1fWR:-Oa-faO56cmeVvf_IF7q:b-a7f0sr) ([raw file](https://github.com/mullvad/pfctl-rs/blob/c3f11c2dde7f7a9d8779d1a2ab40df37550a20cc/src/rule/interface.rs#L68)):*
 > note that conversion from `i8` to `u8` is in theory fallible.

No, casting between integers of the same size is a no-op.
https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.int-same-size

> I don't think `unsafe` itself is a problem.

I respectfully disagree, I'm strongly opposed to unsafe where it's not needed, especially as a reviewer since reviewing unsafe code takes significantly more effort.

> `CString::from_vec_with_nul` will error if `0` byte is not the last one.

Ah, true. I assumed it would truncate the vec.

I would prefer something like this, then:
```rust
let name = pfik_name.map(|b: c_char| b as u8);
let name = CStr::from_bytes_until_nul(&name)
    .map_err(/* etc */)
```

I know that using `map`ing an array for a type cast is a bit ugly. I've been toying with the idea to bring in `zerocopy` for this, as it could replace most uses of `unsafe` we have.

> I have raised this ergonomics issue in the Rust repository

Nice. It would be great to have CStr-methods that dealt with slices of `c_char`.

hulthe (approved):
> @hulthe reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/pfctl-rs/129)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

